PR #23, opened 2014-10-09
Support OrthoXML output for gene trees

muffato:
I have basically copied all the PhyloXML bits and renamed to OrthoXML.
Tested on farm3


muffato:
Still no test-database, but at least the code has been rebased onto a recent head

magaliruffier:
I'll send any complaints to you if something goes wrong because we could not test it ;-)

PR #34, opened 2015-01-15
Updated documentation to show where paramaters go

Relequestual:
It's not clear in the documentation that the paramaters are used in different ways. The required paramater is part of the URL, while the optional paramaters must be part of the message root object. I found this out myself, as I followed the previous example of a single vep query (where the optional paramaters were appened as a query string).

I don't know if I quite got the format your documentation right. Please let me know / fix it if not.

### DIFF
--- a/root/documentation/vep.conf
+++ b/root/documentation/vep.conf
@@ -246,6 +246,7 @@
     output=json
     output=xml
     <params>
+      description=The required paramater is part of the request URL, while the optional paramaters must be included at the root level of the message object.
       <species>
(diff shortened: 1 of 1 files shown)


nerdstrike:
OK, we'll treat this as a bug for now. I'll get back to you.

magaliruffier:
Thanks for reporting this back to us, it always nice to see the service is being used.
This was indeed a bug and has now been fixed on the live server.
Do let us know if that solves the issue for you.

PR #35, opened 2015-02-24
Make sure long lookup happens if @captures first element is false

nicklangridge:
Consistent with the condition in the error reporting at line 229


nerdstrike:
Looks sound, thanks!

PR #43, opened 2015-03-23
Use correct logic to check if we got any captures

nicklangridge:


lib/EnsEMBL/REST/Model/Lookup.pm
```diff
@@ -219,7 +219,7 @@ sub find_object_location {
     my $lookup = $c->model($model_name);
     @captures = $lookup->find_object_location($id, $object_type, $db_type, $species, $use_archive);
     #Check if we any conntent or if the 1st element was false (both mean force a long lookup)
-    unless (@captures && $captures[0]) {
+    unless (@captures || $captures[0]) {
```
nerdstrike: This seems to suggest the possibility of there not being a first element. Is this the case?

nicklangridge:
I believe so; if the lookup finds no hits. This seems consistent with the original condition
if(! @captures) {...} which worked, but failed to check for a falsy first element.

nicklangridge:
You're right, this is not correct, I'll create a new pull request

PR #45, opened 2015-03-31
exit earlier if no data

sarahhunt:
Fix to avoid warn message when query region has no data


nerdstrike:
Thanks Sarah, it's a bit late to go live. I suppose we can try to sneak it in later.

sarahhunt:
No problem - this is a non-urgent fix. It doesn't change what a user 
sees but means you don't get a warn in the logs if they query for a 
region with no data.

magaliruffier:
it can go into next release though
thanks for the change

PR #47, opened 2015-05-01
variation API removed empty fields from returned JSON

willmclaren:
Fixing the tests which I have broken again :-)


magaliruffier:
Great, thanks ;-)

PR #49, opened 2015-06-01
fix for fetching regfeats

willmclaren:


nerdstrike:
Looks good.

magaliruffier:
we'll ignore travis failure for the time being, thanks for the fix

PR #50, opened 2015-06-08
addition of topology of seq_region

danstaines:
Added attribute for topology of seq_region in response to user request

### DIFF
--- a/lib/EnsEMBL/REST/Controller/Assembly.pm
+++ b/lib/EnsEMBL/REST/Controller/Assembly.pm
@@ -53,12 +53,14 @@ sub seq_region: Chained('species') PathPart('') Args(1) ActionClass('REST') {
   my $include_bands = $c->request->param('bands') || 0;
   my $slice = $c->model('Lookup')->find_slice($name);
   my $bands = $c->model('Assembly')->get_karyotype_info($slice) if $include_bands;
+  my $topology = $slice->is_circular()==1?'circular':'linear';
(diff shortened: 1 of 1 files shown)


magaliruffier:
we already have a 'is_circular' flag on some endpoints (overlap and lookup I believe), can we use the same attribute or does it have to be called topology?
also, the assembly_info test case needs updating to reflect that additional field
although travis is not showing this because of the missing JSON library, which I cannot reproduce locally
I know Json.pm is used for the eg lookup, do we need to add it in the cpan file?

danstaines:
On 09/06/15 09:56, Magali Ruffier wrote:
> we already have a 'is_circular' flag on some endpoints (overlap and
> lookup I believe), can we use the same attribute or does it have to be
> called topology?

Sure, consistency makes sense - I thought topology was more obvious, but 
as long as the information is there, thats all I'm after...

-- 
Dan Staines, PhD               Ensembl Genomes Technical Coordinator
EMBL-EBI                       Tel: +44-(0)1223-492507
Wellcome Trust Genome Campus   Fax: +44-(0)1223-494468
Cambridge CB10 1SD, UK         http://www.ensemblgenomes.org/

magaliruffier:
the Json dependency error has been fixed, the CHI library was updated to use a different JSON library instead and we were not importing it any more
I will close this ticket and let you re-submit a request with updated is_circular flag and updated test case

PR #52, opened 2015-06-17
added is_circular attribute for topology of seq_regions

danstaines:
Addition of topology information, using is_circular rather than topology for consistency with other endpoints.


magaliruffier:
thanks for keeping Travis and myself happy

PR #53, opened 2015-06-19
GA4GH endpoints updated after sample schema change

sarahhunt:
plus minor allele addition to variation endpoint


magaliruffier:
looks fine to me

PR #55, opened 2015-07-21
Allow mult. users to serve from the same machine

muffato:
This is achieved by adding the username to the path under /tmp. Similar to
bf6bc139 but this time on TT.pm


nerdstrike:
I guess that's sensible.

PR #56, opened 2015-07-21
Accept transcript and translation IDs in the gene-tree endpoint

muffato:


muffato:
RT #474615

magaliruffier:
nice work, more extensive endpoints are always welcome

nerdstrike:
Bonus points for some minimal tests? I guess we can merge it anyway.

muffato:
Unfortunately, we're still missing a test-database for gene-trees and homologies

PR #57, opened 2015-07-21
Added a cpanfile

muffato:
which is basically a copy of Makefile.PL, but it simplifies setting-up a REST server on the user side and in other teams' test-suites

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,9 +32,6 @@ before_install:
 
 install:
     - cpanm -v --installdeps --with-recommends --notest --cpanfile ensembl/cpanfile .
-    - cpanm -n Catalyst::Runtime 
-    - cpanm -n Catalyst::Plugin::ConfigLoader Catalyst::Action::RenderView Mojo Catalyst::Plugin::Static::Simple
-    - cpanm -n Catalyst::Devel
     - cpanm -v --installdeps --with-recommends --notest .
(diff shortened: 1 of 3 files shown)


magaliruffier:
fair enough, can't see any issue with this

andrewyatz:
Only issue I can think is that we've got a duplication of dependencies in Makefile.PL and the cpanfile. Though the later of the two is more useful. Can the Makefile bring in the cpanfile?

Docs should be updated to reflect this as well

Andy

muffato:
Done. Makefile.PL now loads the cpanfile

magaliruffier:
cool, thanks ;-)

andrewyatz:
Is that all that's needed to get a Makefile.PL to use a CPANfile to populate its dependency list? Just I've seen modules like http://search.cpan.org/~miyagawa/Module-Install-CPANfile-0.12/lib/Module/Install/CPANfile.pm that would suggest it's not as simple as a `do` statement. Perhaps I'm just being too worried

magaliruffier:
I tested two empty VMS, one with cpan, the other with Makefile.PL, and seem to be getting to the same end result

andrewyatz:
Ok that’s good. There are some explicit modules for this in the make module catalyst has chosen to use but I’m happy if it continues to build well :)

PR #58, opened 2015-07-23
New option to output the CIGAR line of the alignment

muffato:
cf RT #475008

- New option in the gene-tree end-points to output the cigar-lines. Default is False to preserve the current behaviour
- New option in the homology end-points to output the cigar-lines. Default is True to preserve the current behaviour. I would personally prefer the default to be False but it would mean changing the major version number :/


magaliruffier:
fair enough for the default behaviour, I could see how users would be confused if the alignments disappear
Shame there are no test cases for gene trees, we'll just have to trust you on this

PR #59, opened 2015-07-31
Do not try to use ncbi_taxonomy as an Ensembl db

nicklangridge:


avullo:
Great, this can go in ... thanks

magaliruffier:
as discussed, happy to take the change in

PR #60, opened 2015-07-31
Use default_compara setting from Controller::Compara config if available

nicklangridge:


avullo:
Looks ok to me as well

PR #61, opened 2015-07-31
Release version should be assigned per-species instead of per-database

nicklangridge:


nicklangridge:
The release version was not being set for species in collection dbs and the so a release version of '0' was being returned by the /info/data/ endpoint.

magaliruffier:
thanks for the fix, looks good to me

PR #62, opened 2015-08-04
Update variation.conf

fcunningham:
changed variation to variant in the text

### DIFF
--- a/root/documentation/variation.conf
+++ b/root/documentation/variation.conf
@@ -10,7 +10,7 @@
     <params>
       <id>
         type=String
-        description=Variation id
+        description=Variant id
         example=__VAR(variation_id)__
         required=1
       </id>
--- a/root/documentation/variation.conf
+++ b/root/documentation/variation.conf
@@ -1,7 +1,7 @@
 <endpoints>
   
   <variation_id>
-    description=Uses a variation identifier (e.g. rsID) to return the variation features
+    description=Uses a variant identifier (e.g. rsID) to return the variation features
     endpoint=variation/:species/:id
     method=GET
     group=Variation
@@ -82,7 +82,7 @@
   </variation_id>
 
   <variation_post>
-    description=Uses a list of variation identifiers (e.g. rsID) to return the variation features
+    description=Uses a list of variant identifiers (e.g. rsID) to return the variation features
     endpoint="variation/:species/"
     method=POST
     group=Variation


magaliruffier:
if we want to be consistent, it probably needs to be changed for the other endpoints (VEP, GA4GH, overlap) as well as in the name space

fcunningham:
Hi Mag,

Ahh good point!

For this page http://rest.ensembl.org/documentation/info/variation_id
Where can I edit this?

I think can leave the title as 'Variation' and the object name as
'variation' e.g. 'variation features'. (I know it isn't perfect).
I changed vep.conf and overlap too.

Fiona
-----------------------------------------------------------------
Fiona Cunningham, PhD
Variation Annotation Coordinator,
European Bioinformatics Institute (EMBL-EBI)
www.ensembl.org || www.lrg-sequence.org

magaliruffier:
the variation.conf file should have all the mentions of variation from the http://rest.ensembl.org/documentation/info/variation_id page
https://github.com/fcunningham/ensembl-rest/blob/patch-1/root/documentation/variation.conf#L3
https://github.com/fcunningham/ensembl-rest/blob/patch-1/root/documentation/variation.conf#L13
...
unless you're talking about changing the url as well? I would rather avoid that if we can, as it would require all our users to update their scripts

fcunningham:
Ok I think I did everything but let me know if you need more. No, no, no,
to changing the URLs ;o)

Fiona
-----------------------------------------------------------------
Fiona Cunningham, PhD
Variation Annotation Coordinator,
European Bioinformatics Institute (EMBL-EBI)
www.ensembl.org || www.lrg-sequence.org

magaliruffier:
What about the example parameters?
id	String	Variation id	-	rs56116432
Do you want to change it to Variant id?
It's line 13 in this file
https://github.com/fcunningham/ensembl-rest/blob/patch-1/root/documentation/variation.conf#L13

fcunningham:
Yes that would make sense. It just looked a bit like code and i was scared
I would break something. Do you want me to do it or will you?

Fiona
-----------------------------------------------------------------
Fiona Cunningham, PhD
Variation Annotation Coordinator,
European Bioinformatics Institute (EMBL-EBI)
www.ensembl.org || www.lrg-sequence.org

magaliruffier:
if you can add this to this pull request, we can have all the changes in one place

fcunningham:
Ok done.

Fiona
-----------------------------------------------------------------
Fiona Cunningham, PhD
Variation Annotation Coordinator,
European Bioinformatics Institute (EMBL-EBI)
www.ensembl.org || www.lrg-sequence.org

magaliruffier:
looks good, thank your

PR #66, opened 2015-08-05
add missing filter

sarahhunt:


magaliruffier:
tests fail because of mismatching release versions and slight alterations to BED format
otherwise the change is fine

PR #67, opened 2015-08-06
dont construct allele string for SV types

willmclaren:
SV types were having an allele string constructed for them, which mean they weren't getting parsed into StructuralVariationFeatures as intended.

This also meant time was wasted looking up sequence.


magaliruffier:
if we're happy this fixed it on live, we can also pull this onto master

willmclaren:
Yeah go ahead - I think I already opened a PR on master?

PR #69, opened 2015-08-10
Patched the Compara database to the e82 schema

muffato:


muffato:
The first build failed with many occurrences of "DBD::mysql::db do failed: Specified key was too long". This was probably due to the fact that the dump script removed all the "DEFAULT CHARSET=latin1" instructions. I've now put them back manually

muffato:
:v thank you mag for the tip !

magaliruffier:
thanks for the patch
I will investigate why the charset gets lost sometimes

muffato:
I ran this on compara3: `perl ../ensembl/misc-scripts/db/dump_mysql.pl --host=compara3 --port=3306 --user=ensadmin --pass=${ENSADMIN_PSW} --database mm14_test_db_multi_compara_20150810_104741 --testcompatible --directory /tmp`
The database is still there

magaliruffier:
the --testcompatible option removes all the special table options
https://github.com/Ensembl/ensembl/blob/release/81/misc-scripts/db/dump_mysql.pl#L329
it sounds like this was added to help with the diff
I need to test whether we can leave some of those options in
Have you tried running the ensembl-test/scripts/patch_test_databases.pl script?
It is meant to do the patching+ dumping for you, and does not strip out the charset

muffato:
On ensembl-compara, `patch_test_databases.pl` worked fine and didn't strip the `CHARSET` stuff.
But on ensembl-rest,  it refused to create the databases (see below) so I felt back to the DIY approach, especially because I only had to patch the compara db.

OK, I've tried again  `patch_test_databases.pl` and I think I understand the error message: the problem is that I was running it from my laptop (usually I run it on farm3), so the network configuration is dodgy

nerdstrike:
Travis is happy, I'm happy.

PR #70, opened 2015-08-25
Add code to support use of VEP plugins

willmclaren:


magaliruffier:
is there no config for plugins in the main config file?
or will that be added as and when we enable plugins?

willmclaren:
The config file to build from is available from the plugins git repo, https://github.com/ensembl-variation/VEP_plugins/blob/master/plugin_config.txt

For most plugins it will need editing before going into production e.g. to point to the location of data files.

lib/EnsEMBL/REST/Controller/VEP.pm
```diff
@@ -438,9 +449,95 @@ sub _include_user_params {
...
+  # get config from file
+  my $plugin_config_file = $vep_config->{plugin_config};
+  return unless $plugin_config_file && -e $plugin_config_file;
+
+  open IN, $plugin_config_file or return;
+  my @content = <IN>;
+  close IN;
+  
+  my $VEP_PLUGIN_CONFIG = eval join('', @content);
+  $c->log->warn("Could not eval VEP plugin config file: $@\n") if $@;
```
avullo: Would add here "and return if ..." to avoid the evaluation of the following statements.

PR #71, opened 2015-08-25
Add a new endpoint /info/variation/:species, providing the list of va…

ens-lgil:
…riation sources for a species

### DIFF
--- a/t/info.t
+++ b/t/info.t
@@ -163,7 +163,7 @@ is_json_GET(
 #/info/variation/:species
 {
   my $variation_json = json_GET('/info/variation/homo_sapiens', 'Get analysis hash');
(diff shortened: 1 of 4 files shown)


t/info.t
```diff
@@ -160,4 +160,26 @@ is_json_GET(
   );
 }
 
+#/info/variation/:species
+{
+  my $variation_json = json_GET('/info/variation/homo_sapiens', 'Get analysis hash');
+  cmp_ok(scalar(keys %{$variation_json}), '==', 6, 'Ensuring we have the right number of sources available');
```
magaliruffier: I think this should be cmp_ok(scalar(@{$variation_json})
Message from travis: Not a HASH reference at t/info.t line 166.

PR #72, opened 2015-08-26
Add a new endpoint /info/variation/:species, providing the list of va…

ens-lgil:
…riation sources for a species.
Should be ok with the tests now.

### DIFF
--- a/lib/EnsEMBL/REST/Controller/Info.pm
+++ b/lib/EnsEMBL/REST/Controller/Info.pm
@@ -177,6 +177,24 @@ sub species_sets : Chained('/') PathPart("info/compara/species_sets") Args(1) Ac
 
 }
 
+sub variation :Local :ActionClass('REST') :Args(1) { }
+
+sub variation_GET :Local :Args(1) {
+  my ($self, $c, $species) = @_;
+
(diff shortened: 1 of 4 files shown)


magaliruffier:
Looking good, thanks for the new endpoint

nerdstrike:
A couple of stylistic observations that don't affect function, or the merging of this new endpoint: 

Passing of a parameter as a method argument, but species via the stash is curious but not harmful. The stash only really comes in useful when you're chaining endpoints together.

Using Args() on the empty method handler, then on the _GET variant is a little odd. I've found Catalyst/Catalyst::Action::REST to be quite delicate to misuse of Args() and CaptureArgs(), so caution is wise.

ens-lgil:
Hi Kieron,

Concerning the species via the stash, I wanted to be consistent with the way the Variation model get the species information. I agree that the other solution would have been to be pass the species data as a method argument.

For the Args(), I have to confess I copied what was done with the other endpoints with the same kind of construction "/info/<type>/:species".
Feel free to edit the Info Controller.

nerdstrike:
If we're going to break the cargo cult, we should do it concertedly. I'll make a ticket for us to tidy this lot up at some point in the future.

PR #73, opened 2015-08-27
Overlap endpoint update

sarahhunt:
Add filter by variant_set to overlap region and id endpoints. Return clinical significance for all three overlap endpoints


magaliruffier:
nice work, I'm sure Simon will be pleased
For reference, we had a ticket for it here: https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-1445

simonbrent:
Looks like what I was after. Where can I test it?

nerdstrike:
Start your own server.

Or wait until release 83. Our testing server goes up one week beforehand.

simonbrent:
So I'm testing this now and I have a suggestion...for it to be easy to use you need to supply a list of what variant_set can be (effectively the result of select name, value from variation_set vs, attrib a where a.attrib_id = vs.short_name_attrib_id), since these values are non-trivial to guess in many cases.

It took me about 5 minutes to work out what that query should be, so I'd have to assume that the process would be much slower for someone without my level of familiarity with the Ensembl code and database.

sarahhunt:
Good point - thanks Simon!

PR #74, opened 2015-09-08
Update the Variation test, because of some changes in the Ensembl Var…

ens-lgil:
…iation API


magaliruffier:
it is just the lower casing, isn't it?
can't see any problem with this, thanks for patching

ens-lgil:
Actually it uses to be the "display term" (i.e. upper case and space in the examples) but the "SO term" should be the default term (i.e. lower case and under score in the examples).
We just changed the Ensembl Variation API to return the SO term by default.

avullo:
Sorry, maybe it's just me but I fail to see the differences between the proposed and current version.

ens-lgil:
The minor changes concern the "most_severe_consequence" attribute:
- First change: 'Intron variant' becomes 'intron_variant'
- Second change: 'Intergenic variant' becomes 'intergenic_variant'

avullo:
Thanks Laurent, a coffee this morning was not enough :)

PR #75, opened 2015-09-08
updates to the GA4GH endpoints for e!82

sarahhunt:


magaliruffier:
it's been rebased against master with the latest patches, looking good now

ensembl_rest_testing.conf
```diff
@@ -129,6 +129,7 @@ jsonp=1
 
 <Model::ga4gh::variantSet>
    ga_config = __path_to(t/test-genome-DBs/testdata/ga_data/ga_vcf_config.json)__
+   dir       = __path_to(t/test-genome-DBs/testdata/ga_data/)__
```
avullo: There's a lot of redundancies in this part (i.e. Model::ga4gh) of the configuration file. All the subsections refer to the same ga_config file and their directories point to the same location.
I haven't seen the production configuration file, are all these to be deemed potentially different on production? If not, they could be shared up to the higher level.

lib/EnsEMBL/REST/Model/ga4gh/variantSet.pm
```diff
@@ -78,23 +78,22 @@ sub fetch_sets{
...
+    ## get info descriptions from one of the VCF files    
+    my $meta = $self->get_info($dataSet);
+
+    ## add summary of essential info for meta for the data set from the config
     foreach my $key ("assembly", "source_name", "source_url"){
       my %meta;
       $meta{key}   = $key;
       $meta{value} = $dataSet->$key;
-      push @meta, \%meta;
+      push @{$meta}, \%meta;
```
avullo: Speeding things up a bit:
map { push @{$meta}, { key => $_, value => $dataSet->$_ } qw ( assembly source_name source_url )

avullo:
Nice, looks good to me as well.

PR #76, opened 2015-09-14
Feature 82/variants in ld

at7:
This PR has two new endpoints: 
- return all variants that are in LD with a given variant
- return list of populations stored in a variation database for a given species (info)

I have added tests for the new endpoints. Some tests required more data in core and variation test databases.


magaliruffier:
the tests are failing because the calc_genotypes binary is missing: travis does not have any pre-set environment
it should be possible to compile it in the variation directory, like we are doing for the tabix executable:
https://github.com/Ensembl/ensembl-rest/blob/release/81/.travis.yml#L21
also, a number of commits seem to be repeated, any chance these could be cleaned up a bit?

PR #77, opened 2015-09-15
Feature 82/final variants in ld

at7:
This PR has two new endpoints:
- return all variants that are in LD with a given variant
- return list of populations stored in a variation database for a given species (info)

I have added tests for the new endpoints. Some tests required more data in core and variation test databases.

And a clean commit history!


magaliruffier:
it is looking good now, thanks for the changes and the test cases

lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
```diff
@@ -0,0 +1,104 @@
...
+    my $pa = $c->model('Registry')->get_adaptor($species, 'Variation', 'Population');     
+    my $population = $pa->fetch_by_name($population_name);
+    if (!$population) {
+      Catalyst::Exception->throw("Could not fetch population object for population name: $population_name");
+    }
+    my $ldfc;
+    try {
+      $ldfc = $ldfca->fetch_by_VariationFeature($vf, $population);
+    } catch {
+      warn "caught error: $_";
```
avullo: Could use the logger here if the warning is of interest knowing.

PR #78, opened 2015-09-21
add doc elements for VEP plugins

willmclaren:
These are the plugins I intend to have live this release.

I'll create a PR to the private repo with the configs


magaliruffier:
that's going to make for a busy documentation page, but I think it is a necessary evil

PR #79, opened 2015-09-22
Replace Bio::DB::Fasta with faidx_xs

willmclaren:
Benefits:

- can index bgzipped FASTA files (though note Ensembl distributes gzipped, must be converted to bgzipped before use). Human toplevel file is ~1.1g bgzipped, 40g uncompressed.
- indexes much more robust, won't tolerate corrupted/partial index like Bio::DB::Fasta
- around 10x faster retrieving sequence

Requires:
- htslib (https://github.com/samtools/htslib)
- faidx_xs (https://github.com/Ensembl/faidx_xs)
- bgzipped FASTA file

The .travis.yml install of htslib and faidx_xs is non-ideal as it avoids using "make install" by compiling the module into the home directory. Much better if you have sudo access to do "make install" properly for each.


nerdstrike:
Looks sensible.

avullo:
Same for me, OK!

PR #80, opened 2015-09-23
Feature/ld documentation

at7:
Update parameter names in ld.conf


avullo:
Great, changes can go in.

PR #83, opened 2015-09-25
add variant set list

sarahhunt:
Improve overlap documentation as per Simon's suggestion.


magaliruffier:
thanks for the change, that will be really useful

PR #85, opened 2015-09-30
fetch compara adaptors, used by plugins

willmclaren:


willmclaren:
This will also need a git pull in the VEP_plugins directory to work

magaliruffier:
I have updated the test rest with the changes, but I still can't see any conservation score
the message about missing host is gone though
I am also slightly worried that the tests are failing, something about a 'Multi' species being passed around with the VEP endpoint?

muffato:
Perhaps instead of using multi you could call get_best_compara_DBAdaptor with the current species name: https://github.com/Ensembl/ensembl-rest/blob/release/81/lib/EnsEMBL/REST/Controller/Homology.pm#L50
This way it will also work for plants etc

PR #86, opened 2015-09-30
Fixes so Conservation plugin works

willmclaren:


magaliruffier:
I've checked out the branch on the test starman, that seems to fix it:
http://rest.ensembl.org:8080/vep/human/id/rs699?content-type=application/json&Conservation=1&Blosum62=1
Also, travis is happy now :-)

PR #88, opened 2015-10-08
Correct summary_as_hash key

sarahhunt:
The name give to the allele string in the VariationFeature summary_as_hash method is missleading, so has been updated. This updates the relevant REST test.

### DIFF
--- a/t/overlap.t
+++ b/t/overlap.t
@@ -201,7 +201,7 @@ my $base = '/overlap/region/homo_sapiens';
     consequence_type => 'intergenic_variant',
     feature_type => 'variation',
     seq_region_name => '6',
-    alt_alleles => [ qw/T -/]
+    alleles => [ qw/T -/]
   }, 'Checking one variation format');
   
   my $somatic_json = json_GET("$base/$region?feature=somatic_variation", 'Fetching somatic variations at '.$region);


nerdstrike:
A seismic change in the code base. Are you sure?

magaliruffier:
change will need to go in, but a big version change is now required

muffato:
if there is the opportunity, I would probably find fields to rename in Compara too. But perhaps we could delay the major version bump until we have more things to change than just field names ?
Can the method return both "alleles" and "alt_alleles" for the time being ?

sarahhunt:
I assume when the key was created it was intended to mean 'all possible 
alleles at a location', but it is being interpreted as 'all 
non-reference alleles at a location'. The former is the information most 
people will be seeking. We don't have a quick enough way of identifying 
alt_alleles to add to this endpoint unfortunately.

Documentation of what the fields mean would help here.

magaliruffier:
even if we could correctly populate the "alt_allele" field, that would be even worse, as we would be replacing the expected data in an existing field
And providing both "alleles" and "alt_alleles" as duplicate will only add more confusion before fully correct this
It is not ideal, but I'd rather make the brutal change now and save more confusion
If compara has some field renaming, we will just bump up the version again next release, and we will be on version 10 of the REST API in no time!

PR #91, opened 2015-10-23
ENSCORESW-1480

lairdm:
- Allow users to specify start and end coordinates for a sequence when using the /sequence/id endpoint
- Testing for sub-sequence calls
- Added new tests to harness to check return code and do a regex test on the response regardless of return code (ie. text response on a 400)
- Modified tests to use these new test types for start/end coordinates beyond sequence boundaries

### DIFF
--- a/lib/EnsEMBL/REST/Controller/Sequence.pm
+++ b/lib/EnsEMBL/REST/Controller/Sequence.pm
@@ -156,6 +156,13 @@ sub _process {
   Catalyst::Exception->throw(qq{The type '$type' is not understood by this service}) unless $allowed_values{type}{$type};
   
   if($c->request->param('start') || $c->request->param('end')) {
+      # We don't support circular features at this time, if it's a circular
+      # feature, and not a translation, throw an error.
+      unless($object->isa('Bio::EnsEMBL::Translation')) {
+	  my $feat_slice = $object->feature_Slice();
+	  Catalyst::Exception->throw(qq{Circular slices aren't supported for sequence trimming at this time}) if($feat_slice->is_circular());
+      }
+
       # It doesn't make sense to expand the sequence when doing a sub-sequence
       if($c->request->param('expand_5prime') || $c->request->param('expand_3prime')) {
 	  Catalyst::Exception->throw(qq{You may not expand the 3prime or 5prime sequence end when requesting a sub-sequence});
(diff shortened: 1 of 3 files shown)


lib/EnsEMBL/REST/Controller/Sequence.pm
```diff
@@ -188,39 +198,60 @@ sub _process_feature {
...
     $seq = $object->transcript()->translate()->seq();
+    # Grab the subsequence if we've been asked to trim the ends
+    $seq = $self->_process_subseq($c, $seq) if($c->stash()->{dosubseq});
   }
   #Transcripts
   elsif($object->isa('Bio::EnsEMBL::PredictionTranscript')) {
     if($type eq 'cdna') {
       $seq = $object->spliced_seq($mask_feature);
+      # Grab the subsequence if we've been asked to trim the ends
+      $seq = $self->_process_subseq($c, $seq) if($c->stash()->{dosubseq});
```
nerdstrike: Abundant copy-paste implies there might be room for simplification here, with only a single _process_subseq() call for any of the $seq outputs. Clarity should take priority if the subseq can be easily overlooked.

lib/EnsEMBL/REST/Controller/Sequence.pm
```diff
@@ -265,6 +308,119 @@ sub _process_feature {
...
+
+  # Sanity checking
+  if($start < 0 || $start > $seq_len) {
+      Catalyst::Exception->throw(qq{Your start coordinate is not within the sequence requested})
+  }
+  if($end < 1 || $end > $seq_len + 1) {
+      Catalyst::Exception->throw(qq{Your end coordinate is not within the sequence requested})
+  }
+  if($start > $end) {
+      Catalyst::Exception->throw(qq{Your start coordinate cannot be larger than your end})
```
nerdstrike: Circular genomes applicable here @avullo ?

lib/EnsEMBL/REST/Controller/Sequence.pm
```diff
@@ -265,6 +308,119 @@ sub _process_feature {
...
+  if($pep_end < $pep_start) {
+      $c->request->params->{'start'} = 0;
+      $c->request->params->{'end'} = 0;
+  } else {
+      $c->request->params->{'start'} = $pep_start;
+      $c->request->params->{'end'} = $pep_end;
+  }
+
+  # Remember we've translated the coordinates so we don't try again
+  $c->{stash}->{coordstranslated} = 1;
```
nerdstrike: Direct access of stash hash is a bit naughty. These should be set via the accessor. $c->stash->{coordstranslated} = 1;

t/sequences.t
```diff
@@ -266,7 +266,112 @@ FASTA
...
+  action_check_code($url, 400, 'Return code for out of boundaries parameters should be 400');
+  action_raw_bad_regex($url, qr/not within the sequence/, 'Start of sub-sequence range can not be beyond the sequence length');
+}
+{
+  my $url = "/sequence/id/?start=120&end=150";
+  my $body = q/{ "ids" : [ "ENSP00000370194", "ENSG00000243439" ]}/;
+  my $seq_a = q/LAEYTADVDGVGTLRLLDAVKTCGLINSVKF/;
+  my $seq_b = q/TGCACTAAGTTCAGCATGAAGAGCAGCGGGC/;
+  my $response = [{"desc" => undef,"id" => "ENSP00000370194","seq" => $seq_a,"molecule" => "protein"},{"desc" => "chromosome:GRCh37:6:1507676:1507706:1","id" => "ENSG00000243439","seq" => $seq_b,"molecule" => "dna"}];
+  is_json_POST($url,$body,$response,'POST ID sequence fetch with sequence trimming');
```
nerdstrike: I'd like to see a test that probes the behaviour of a start = end subseq, for a single base.

lib/EnsEMBL/REST/Controller/Sequence.pm
avullo: Yes, it should take into account circular genomes as well, in principle. My work on them in the past was mainly about making sure the features crossing the origin had the correct coordinates wrt their associated slice. I didn't thoroughly test the sequence aspect of circular genomes. I'd leave Matthew constraint at the moment, and see if EG is interested in the exception.

lib/EnsEMBL/REST/Controller/Sequence.pm
lairdm: Fair enough...

lairdm:
Pushed fixes based on feedback received.

nerdstrike:
I think it's good to go.

PR #92, opened 2015-11-05
Enscoresw 1551

lairdm:
Rate limit override headers to disable 429 return code
Tests to demonstrate the override works

### DIFF
--- a/lib/Plack/Middleware/EnsThrottle.pm
+++ b/lib/Plack/Middleware/EnsThrottle.pm
@@ -45,6 +45,8 @@ enable 'EnsThrottle::Second',
   },
   blacklist => '192.168.2.1', #blacklist a single IP
   whitelist => ['192.168.0.0-192.169.255.255'], #whitelist a range
+  whitelist_hdr => 'Token',
+  whitelist_hdr_values => ['loveme'],
   client_id_prefix => 'second', #make the generated key more unique
   message => 'custom rate exceeded message',
   # If specified once limit has been hit add this value to the Retry-After header.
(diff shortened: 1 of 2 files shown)


lib/Plack/Middleware/EnsThrottle.pm
```diff
@@ -45,6 +45,8 @@ enable 'EnsThrottle::Second',
   },
   blacklist => '192.168.2.1', #blacklist a single IP
   whitelist => ['192.168.0.0-192.169.255.255'], #whitelist a range
+  whitelist_hdr => 'Token',
+  whitelist_hdr_values => ['loveme'],
```
nerdstrike: My concern here is we are publishing the unlimited token on Github.

lib/Plack/Middleware/EnsThrottle.pm
```diff
@@ -45,6 +45,8 @@ enable 'EnsThrottle::Second',
   },
   blacklist => '192.168.2.1', #blacklist a single IP
   whitelist => ['192.168.0.0-192.169.255.255'], #whitelist a range
+  whitelist_hdr => 'Token',
```
andrewyatz: Token isn't an official HTTP header. Perhaps it's better to use something more explicit like `X-RateLimit-Token` as the example header.

lairdm: That's an example and in the documentation section, the "live" config is of course hidden in the private github. I could change it to something more accurate, at the same time I'm not sure I want anything close to what the "real" value we use will be :)  "X-My-Secret-Header"?

PR #93, opened 2015-11-13
Enabling pre-load of the Registry

andrewyatz:
Object uses an after BUILD modifier to trigger the population of the
registry and associated lookups. Use with a pre-fork server to
avoid nasty running server process startup costs caused by reaping
processes.

### DIFF
--- a/ensembl_rest.conf.default
+++ b/ensembl_rest.conf.default
@@ -20,15 +20,15 @@ jsonp=1
   # user = anonymous
 
   host = ensembldb.ensembl.org
-  port = 5306
+  port = 3306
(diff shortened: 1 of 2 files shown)


andrewyatz:
Please check this

muffato:
Interesting. I may add a patch to load the Compara basic objects too

andrewyatz:
Ok sounds good. I assume you're talking about GenomeDBAdaptor & MLSS?

muffato:
I am

magaliruffier:
should those changes also go into the production ensembl_rest.conf.default?

PR #94, opened 2015-11-30
Test-database cleanup

muffato:
Some tables are not needed


magaliruffier:
can't complain about making the tests simpler, the change is welcome
is it ok if we wait a few days before merging it in though?
We've just branched, and I'd rather not have master and branch diverge so soon

PR #95, opened 2015-11-30
Several improvements for the Homology endpoints

muffato:
In short:
 - the Compara API is able to filter the target species
 - a few things could be rewritten in a clearer way


magaliruffier:
that looks fine to me
Am I right in understanding that this does not affect the current behaviour of the endpoint, but makes the code behind it a bit more efficient?
Same caveat as for the other PRs, can we wait a few days before merging?

muffato:
You're right: this is not a major version-number increase

magaliruffier:
were you planning to get those changes in for release 83?

muffato:
I think it's too late for that

magaliruffier:
agreed, just making sure :-) Thanks for the changes either way!

PR #96, opened 2015-11-30
Several improvements to the Gene-tree endpoints

muffato:
In short:
 - added options to prune the gene-trees (like on the homology side)
 - added more examples

### DIFF
--- a/ensembl_rest.conf.default
+++ b/ensembl_rest.conf.default
@@ -112,6 +112,7 @@ jsonp=1
     taxon_name=Homo sapiens
     taxon_nameish=Homo%25
     target_taxon=10090
+    target_ancestral_taxon=9526
     target_species=cow
     
     ontology=GO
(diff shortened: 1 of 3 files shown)


magaliruffier:
can I have some basic tests for it as well?
If I understand correctly, this will not affect existing output but allows for two new options, is that right?
we'll declare these for release 84 and add a minor version increment

muffato:
Tests ... Perhaps in e84 ? That's the reason why this PR and #95 have been pending for a couple of releases. I was hoping to find a way of making a Compara DB but I haven't yet

magaliruffier:
I take it the test database we have does not cover enough cases?

muffato:
There is nothing related to gene-trees / homologies in this test-db

magaliruffier:
I can see that being a problem

muffato:
It is, I know

PR #99, opened 2015-12-10
Speed up transcript_variation fetch (copy of master PR)

willmclaren:
Use a hash instead of grepping loads, and assign the VF
properly to the TV so it doesn't go and load it again.

Also needs a fix to VariationFeatureOverlap.pm in
ensembl-variation to properly go faster.

### DIFF
--- a/lib/EnsEMBL/REST/Model/Overlap.pm
+++ b/lib/EnsEMBL/REST/Model/Overlap.pm
@@ -293,7 +293,7 @@ sub somatic_transcript_variation {
 sub _filter_transcript_variation {
   my ($self, $transcript_variants) = @_;
   my $type = $self->context->request->parameters->{type};
-  my $cached_vfs = $self->context->stash->{_cached_vfs};
+  my %cached_vfs = map {$_->dbID => $_} @{$self->context->stash->{_cached_vfs} || []};
   my @vfs;
 
   foreach my $tv (@{$transcript_variants}) {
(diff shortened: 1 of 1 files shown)


andrewyatz:
+1 to removing the linear scan into a hash lookup

willmclaren:
The fix to VariationFeatureOverlap.pm is now in too https://github.com/Ensembl/ensembl-variation/commit/e8e697f97d747b50fc8dda6befba7683b000defd

willmclaren:
And just for some context, this now returns in 4s on my sandbox; it times out on rest.ensembl.org

andrewyatz:
:joy: that's just fantastic

magaliruffier:
Thanks Will, that looks great :-)

nerdstrike:
Great speed-up, brave use of cached_vfs in scalar, list and hash contexts!

willmclaren:
I think I just changed it from an arrayref to a hash, no?

nerdstrike:
It's not a reason not to merge the code, but readability takes a hit.

willmclaren:
Sorry Kieron, you've lost me here.

There is no @$vfs in the method that has been changed (_filter_transcript_variation)

@vfs is an array that gets EnsEMBL::REST::EnsemblModel::TranscriptVariation objects pushed onto it in the loop.

It should probably be called @tvs, but I didn't write this method in the first place so not really up to me what the variables are named.

willmclaren:
I also added an equivalent on master, you may as well merge that too...

magaliruffier:
that's the plan ;-)

PR #100, opened 2016-01-05
Don't use split when retrieving by stable_id for archived identifiers…

lairdm:
…, the ensembl library knows how to handle this now

Also, pass along the determined type of identifier to the Archive Adaptor since some identifiers can't be determined by regex (ie. C. elegans)


magaliruffier:
yeah to cleaner code :-)

nerdstrike:
Do I understand correctly that this will now make two DB calls for every versioned ID? It may be ok anyway I suppose.

PR #104, opened 2016-01-26
uniquify phenotypes before returning

willmclaren:


magaliruffier:
is that to return a single phenotype when there are haplotypes?
looks reasonable to me

PR #105, opened 2016-01-29
use modifier to speed up LD calls

willmclaren:
This modifier tells the variation API to not fetch VariationFeature objects, but rather rely on the position->ID mapping extracted from the VCF during genotype extraction. This should make these REST calls many times faster (5-8x faster in my tests).

It depends on a set of changes in ensembl-variation master for the speedup, but safely falls back to the old behaviour when using release/83.


nerdstrike:
Since I have to say something, it's fine. You could be more defensive about the data coming from the LD container.

nerdstrike:
Also, unit test needs an orderless comparison.

willmclaren:
This passes now...

magaliruffier:
Thanks for the update, looks fine to me

PR #106, opened 2016-02-05
Feature 84/ld increase window

at7:
We have worked on speed improvements for the LD calculation this release. We can now increase the window_size for the LD REST endpoint and make the window_size parameter customisable. Max allowed value is 1MB.

### DIFF
--- a/lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
+++ b/lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
@@ -17,8 +17,7 @@ package EnsEMBL::REST::Model::LDFeatureContainer;
 use Moose;
 use namespace::autoclean;
 use Try::Tiny;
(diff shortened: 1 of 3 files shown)


nerdstrike:
Max value is 10^6 or 10^5?

at7:
1 Mb

nerdstrike:
1000/2 = 500 -> 500 * 100 = 50000. How many windows do you run? Also, I hope the called code can handle 499.5 in a window size.

at7:
You specify the window size in kb: e.g. 25, 500, 1000. The code translates it into a number that's why * 100. 

Thanks for pointing out the division problem. Will add another test.

at7:
Sorry, missing one zero here. I have to multiply by 1000.

nerdstrike:
Not sure why the rate limit test is failing. That's not your fault.

PR #108, opened 2016-02-09
Feature 84/new ld endpoints

at7:
Add new LD endpoint. Compute LD values for a specified region.

### DIFF
--- a/lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
+++ b/lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
@@ -42,9 +42,9 @@ sub fetch_LDFeatureContainer_variation_name {
   my $ldfca = $c->model('Registry')->get_adaptor($species, 'Variation', 'LDFeatureContainer');
 
   my $window_size = $c->request->param('window_size') || 1000; # default is 1MB
-  Catalyst::Exception->throw("window_size needs to be a value bewteen 0 and 1000.") if (!looks_like_number($window_size));
-  Catalyst::Exception->throw("window_size needs to be a value bewteen 0 and 1000.") if ($window_size > 1000);
-  Catalyst::Exception->throw("window_size needs to be a value bewteen 0 and 1000.") if ($window_size < 0);
+  Catalyst::Exception->throw("window_size needs to be a value between 0 and 1000.") if (!looks_like_number($window_size));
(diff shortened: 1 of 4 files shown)


lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
```diff
@@ -81,6 +81,41 @@ sub fetch_LDFeatureContainer_variation_name {
...
+  Catalyst::Exception->throw("Specified region is too large. Maximum allowed size for region is 1Mb.") if ($slice->length > 1_000_000);
+  my $c = $self->context();
+  my $species = $c->stash->{species};
+
+  my $ldfca = $c->model('Registry')->get_adaptor($species, 'Variation', 'LDFeatureContainer');
+
+  my $ld_config = $c->config->{'Model::LDFeatureContainer'};
+  if ($ld_config && $ld_config->{use_vcf}) {
+    $ldfca->db->use_vcf($ld_config->{use_vcf});
+    $Bio::EnsEMBL::Variation::DBSQL::VCFCollectionAdaptor::CONFIG_FILE = $ld_config->{vcf_config};
```
nerdstrike: I've questioned this elsewhere, but this kind of direct variable access is not OO and should be avoided if possible. It makes code untrustworthy and harder to test. It would be nicer if the VCFCollectionAdaptor had an accessor for this.

lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
```diff
@@ -81,6 +81,41 @@ sub fetch_LDFeatureContainer_variation_name {
...
+    my $pa = $c->model('Registry')->get_adaptor($species, 'Variation', 'Population');     
+    my $population = $pa->fetch_by_name($population_name);
+    if (!$population) {
+      Catalyst::Exception->throw("Could not fetch population object for population name: $population_name");
+    }
+    my $ldfc;
+    try {
+      $ldfc = $ldfca->fetch_by_Slice($slice, $population);
+    } catch {
+      warn "caught error: $_";
```
nerdstrike: warn() is bypassing our logging framework. Please use $c->log()->warn($message) or just throw a Catalyst Exception for the controller to deal with. Then we can see that the controller ignores certain kinds of errors in the controller.

PR #109, opened 2016-02-10
Add transcript haplotypes endpoint

willmclaren:
This endpoint returns a serialised JSON version of a
TranscriptHaplotypeContainer; this container contains details
of all observed haplotype sequences for a transcript, both the CDS
and translated protein. The haplotypes are based on phased genotype
data, at the moment the only source for this is 1000 genomes.


nerdstrike:
Looks ok, I've not spun up a test server to look at the docs.

magaliruffier:
does the example config file also need updating?
https://github.com/Ensembl/ensembl-rest/blob/release/83/ensembl_rest.conf.default
And if you have access to the ensembl-rest_private repo, can you update the config files there where required?

willmclaren:
It works off the existing VCF config, so as long as that's set up OK (the one that's there for genotypes), it should be good.

There is one update to the default config, but that's just adding an example transcript ID.

magaliruffier:
ok, thanks
we should be able to test in the next couple of weeks once we set up the REST server, so we might get back to you then ;-)

PR #110, opened 2016-02-17
Fixed error message

simonbrent:
s/Include/Expand/ when expand parameter is given but not allowed


magaliruffier:
Oops, well spotted

PR #111, opened 2016-02-19
Feature 84/ld increase window

at7:
This is another endpoint for LD data computation. This endpoint computes LD data for a pair of variants. Sorry for the misleading branch name.


magaliruffier:
while we are on the subject of LD endpoints, @willmclaren was wondering about listing them under the general variation section rather than in their separate one
Has there been any decision one way or the other?

willmclaren:
I think we decided this was down to the core team as you have "ownership" of the REST interface.

We're happy to go along with whatever arrangement you decide, but also happy to participate in any discussion with others about how we move forwards.

My opinion would be we need to concentrate on organisation more the more endpoints there are. Specifically I would also suggest we need more levels, i.e.

- Variation
  - Variation lookup
  - VEP
    - ID
    - region
  - LD
    - pairwise

magaliruffier:
That's fair.
We will be leaving it as is for the time being then but might plan some larger overhaul of the endpoint organisation at some point
PR looks fine though

lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
```diff
@@ -116,11 +116,67 @@ sub fetch_LDFeatureContainer_slice {
...
+    my $ldfc;
+    try {
+      $ldfc = $ldfca->fetch_by_VariationFeatures(\@vfs, $population);
+    } catch {
+      Catalyst::Exception->throw("Something went wrong while fetching from LDFeatureContainerAdaptor");
+    };
+    return $self->to_array($ldfc)
+  }
+  # compute LD for all LD populations
+  my $ld_populations = $pa->fetch_all_LD_Populations;
```
avullo: I think here it would raise a warning the population adaptor is undefined in the default case. By default, population name is unset hence line 147 would not be executed.

at7: I'm defining a population adaptor in line 126. I removed the second assignment. Thanks!

avullo: that's great thanks, ready to go ...

PR #112, opened 2016-02-25
move to Bio::DB::HTS module from faidx_xs

rishidev:
Changes have been made to the VEP to use Faidx (FASTA indexing from HTSlib) from Bio::DB::HTS rather than its own module. These changes are to accompany this.

Relevant pull requests are:
https://github.com/Ensembl/ensembl-variation/pull/25
https://github.com/Ensembl/ensembl-tools/pull/12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: "perl"
 
 perl:
   - "5.14"
-  
+
 sudo: false
-  
+
 addons:
   apt:
     packages:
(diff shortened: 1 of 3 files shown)


magaliruffier:
don't the Bio-HTS library need to be added to the PERL5LIB? This would probably need to go in harness.sh
also, are there instructions somewhere on how to add all those dependencies on the REST server?
the travis instructions do not exactly reflect a normal development environment

rishidev:
Good spot - I've updated the PERL5LIB path. 

Instructions on building are in the README for the project https://github.com/Ensembl/Bio-HTS - let me know if you feel it's missing anything.

magaliruffier:
Thanks, it seems to work now
The test suite is failing due to issues in the regulation schema, so not related to this change.
We'll use the README when we set up the test REST server and will let you know if we have any issues

rishidev:
Thanks

lairdm:
I still think it'd be better to make an entry in PERL5LIB pointing to blib/arch and blib/lib rather than copy the .pm and .so for Faidx to a non-standard location. You can point at an xs based module without doing a system wide install, the perl loader will find it if it's in the PERL5LIB.

magaliruffier:
regulation schema has been patched, travis is now happy
Matt's point about setting up is valid though

rishidev:
Not a problem to implement the longer PERL5LIB setting here (it was an issue to ask the VEP users to do so, which is the setup I copied for here) instead of the file copies. However I'd assumed this had been accepted and I deleted my fork as part of a tidy up without double checking that that it indeed had been accepted. Sorry. I will need to make another fork and a new PR for the final step - I can either redo the changes in this PR from scratch or if you're happy to merge this here I can make just the Travis tweaks in the follow-up PR. Let me know which you'd prefer.

magaliruffier:
we'll take these changes in now

rishidev:
Cheers, will make the further update.

PR #114, opened 2016-03-09
VEP now returns cons in ENST order, instead of random Perl order

willmclaren:


magaliruffier:
thanks for the update, looks fine now

PR #116, opened 2016-03-10
Correctly set the "species_set_group" field

muffato:


magaliruffier:
looks simple enough ;-)
tests are failing because of VEP schema changes which are not in yet, so the PR should be good to go

PR #118, opened 2016-03-23
Use tabix from htslib, build Bio::DB::HTS from CPAN

rishidev:


magaliruffier:
the failures are due to the transcript_haplotypes changes, so it should be good to go

magaliruffier:
bioperl is currently down, so all test suites fail
We're going to wait until Monday to see if it is resolved before taking new changes in

PR #119, opened 2016-04-07
add records for implicit _all population to transcript haplotype output

willmclaren:
This along with fix in ensembl-variation should fix broken test suite


magaliruffier:
it is currently breaking due to missing libraries in htslib
that should be fixed by Rishi's PR (https://github.com/Ensembl/ensembl-rest/pull/118) so we can check this one again once the other fix is in

magaliruffier:
bioperl is currently down, so all test suites fail
We're going to wait until Monday to see if it is resolved before taking new changes in

lairdm:
Bioperl is up, merging.

PR #120, opened 2016-04-14
At7/postrelease84/ld windowsize

at7:
I've changed the default window size to 500 KB. That is still very generous compared to what we use as default on the website. But I hope it will prevent death by LD. I'lll continue looking into how we can speed up the computations.

### DIFF
--- a/lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
+++ b/lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
@@ -41,10 +41,10 @@ sub fetch_LDFeatureContainer_variation_name {
   my $va = $c->model('Registry')->get_adaptor($species, 'Variation', 'Variation');
   my $ldfca = $c->model('Registry')->get_adaptor($species, 'Variation', 'LDFeatureContainer');
 
-  my $window_size = $c->request->param('window_size') || 1000; # default is 1MB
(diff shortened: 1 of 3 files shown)


nerdstrike:
Thanks Anja, Magali has noticed a significant decline in performance versus our test server from release 83. Did a lot change between releases?

at7:
We changed the window size from 50 KB to 1000 KB. We were confident that our speed improvements could deal with this.

magaliruffier:
we're now down to about half the time, 16 seconds rather than 35
If this going to be enough?
btw, the test suite is failing because of the changes to htslib and the like
can you also submit the PR against master where we can check against regressions?

at7:
Can you try any of the requests Kieron has sent me?

/ld/Human/rs1094732?content-type=application/json;population_name=1000GENOMES:phase_3:GBR;r2=0.500000

/ld/Human/rs9302874?content-type=application/json;population_name=1000GENOMES:phase_3:GBR;r2=0.500000

magaliruffier:
the first example one times out on the live server
with the fix, it returns in about 24 seconds
the second example returns in 46 seconds (just misses the time out)
with the fix, results come back in about 21 seconds

at7:
That is much better.

magaliruffier:
ok, if you think those times are reasonable, we can accept the PR and push the changes to the live servers

at7:
yes that's fine. thank you

PR #122, opened 2016-04-15
use fetch_all_by_name_list instead of single fetches

willmclaren:
Might solve your slow ID request problem.

But then again it might not.

It should at least be faster.


nerdstrike:
Faster is good.

magaliruffier:
thanks, I might run that on the test server

PR #123, opened 2016-04-19
EQTL Endpoint

juettemann:


nerdstrike:
Is the EQTL adaptor visible in the Regulation checkout or is it on a branch?

juettemann:
It is a repository:
https://github.com/Ensembl/ensembl-hdf5

Why is this bloody Preload.pm still hanging around?

magaliruffier:
You might want to update travis.yml and possible cpanfile to list all the dependencies you have
It will make travis happier and ensure we have some form of documentation to update the rest server
also, some test cases ;-)

nerdstrike:
@juettemann don't fixate on Preload. We'll fix that once everything else is in order.

magaliruffier:
if you update your branch, that should update the PR
if we can't get some basic instructions that keep travis happy, we have no chance of installing this on our servers, so that's definitely a requirement
can you create a mini eqtl file for testing?

juettemann:
Creating a small DB is (should) not be a big problem. Installing HDF5 [https://www.hdfgroup.org/HDF5/release/obtain5.html] and then the adaptors takes a quite a while (~60 min) and a reasonable amount of manual config. I am sure I can get the config in yaml, but likely need help from one of you in the process.
Would travis build that everytime, or is it a one time thing?

nerdstrike:
Every time anyone commits, travis runs. It might be worth mocking up the HDF5 adaptor so that you can test the calling code. See Mock::Object

t/MultiTestDB.conf.default
```diff
@@ -6,9 +6,10 @@
 {
   'databases' => {
     'homo_sapiens' => { 
-      'core' => 'Bio::EnsEMBL::DBSQL::DBAdaptor',
+      'core'      => 'Bio::EnsEMBL::DBSQL::DBAdaptor',
       'variation' => 'Bio::EnsEMBL::Variation::DBSQL::DBAdaptor',
-      'funcgen' => 'Bio::EnsEMBL::Funcgen::DBSQL::DBAdaptor',
+      'funcgen'   => 'Bio::EnsEMBL::Funcgen::DBSQL::DBAdaptor',
+      'eqtl'      => 'Bio::EnsEMBL::HDF5::EQTLAdaptor',
```
magaliruffier: I think we need to remove that line now

ensembl_rest_testing.conf
```diff
@@ -12,6 +12,10 @@ jsonp=1
 
 <Model::Registry>
   skip_initation = 1
+ preload = 1
```
magaliruffier: do we still need the preload?
and might be cleaner if we remove the lines that have been commented out

.travis.yml
```diff
@@ -16,6 +16,7 @@ env:
 before_install:
     - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-test.git
     - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl.git
+    - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-hdf5.git
```
magaliruffier: if hdf5 is needed for the tests, it should also be added to the perl5lib in the harness

magaliruffier:
Not quite happy yet:
Can't locate loadable object for module Bio::EnsEMBL::HDF5

magaliruffier:
Victory! We have tests and they pass!
@juettemann if you can just remove the eqtl_dbs and preload options from the example config file, I think we're good to go

dzerbino:
Thanks Mag,  that's terrific!

nerdstrike:
While we're at it, can we change the following error for something more targeted?

{"error":"Stable ID 'ENSG00000122435' not present in the current Ensembl release"}

The ID is in the release, just not available to eQTL code.

dzerbino:
Indeed, it's an edge case. If the HDF5 file is created with a slightly
older version of Ensembl genes (as it was with the test) then it does not
contain all the Ensembl gene IDs for the release.

nerdstrike:
Further to my last, I recommend some logging of eQTL initialisation, e.g.

has 'eqtl_dbs'  => (is => 'ro', isa => 'HashRef[Str]', predicate => 'loaded_eqtl_dbs');

 $self->log->debug("EQTL database files:\n");
 $self->log->debug(Dumper $self->eqtl_dbs);
 if (! $self->loaded_eqtl_dbs) {
    $self->log->error('No EQTL database configured, /eqtl endpoints will fail');
    return {};
 }

dzerbino:
I'm confused, has the PR been merged or not?

What remains to be done?

nerdstrike:
If you want this code merged properly, it'll need some "not implemented on this server" type messages for servers that will never have eQTL data, like our GRCh37 one. Otherwise I believe we should keep this set of endpoints firmly in its own branch which we can keep up to date with our master branch as and when needed.

dzerbino:
OK thanks. I think Mag explained this to me earlier, but I had had
forgotten. Sorry.

nerdstrike:
This branch is now redundant. Code resides in ensemb-rest:feature/eqtl , as Thomas is away and not around to modify this pull request. I recommend reforking from there for future development.

PR #124, opened 2016-04-22
Feature84/ld add more data

at7:
Return more variant attributes in the LD endpoint.


at7:
I couldn't find a way to send the whole branch as a PR

magaliruffier:
ok, I've created a new branch from master, can you please rebase your branch against that?

PR #127, opened 2016-05-19
Updates to match Compara master

muffato:
Without this you'll start seeing Travis failing ...


nerdstrike:
A simple enough change. I assume this will require compara 85 or master to be effective.

muffato:
Yes, master (e85 onwards) only. If I manage to make it right ...

nerdstrike:
Note that the Travis errors here are due to changes in the ontology schema that have left the test DBs behind.

muffato:
The errors are in overlap.t and ontology.t . I don't understand how they're related to my change, though

magaliruffier:
the ontology schema has now been patched but the test suite is still failing
I get the following error for genomic_alignment.t
not ok 7 - EPO alignment, block
#   Failed test 'EPO alignment, block'
#   at /nfs/users/nfs_m/mr6/src/ensembl-rest/t/lib/RestHelper.pm line 47.
Can't use string ("0") as an ARRAY ref while "strict refs" in use at t/genomic_alignment.t line 100.

line 100 is the following:
  my $json = json_GET("/alignment/block/region/$species/$region?species_set_group=birds;method=EPO", "EPO alignment, block");
  is (scalar(@{$json}), 1, "number of alignment blocks");

It does not fail on older versions of the compara checkout

muffato:
:v:

magaliruffier:
Awesome, thanks

muffato:
Great, thank you :) I'll soon make another pull-request with changes in the gene-tree and homology endpoints

PR #128, opened 2016-05-31
Compara performance improvements

muffato:
There are three categories of improvements:

1) Number of queries. I have added several helper methods in the Compara API that call `fetch_all_by_dbID_list()`. This is as you imagine much faster than calling `fetch_by_dbID()` repeatedly (PhyloXML export of a big tree down from 30 sec to 6 sec) and the helper methods are automatically called whenever needed in the Compara API. That's why I could remove a number a `preload()` call from ensembl-rest. Another consequence is that it now takes a constant number of queries to fetch the data for the gene-tree endpoints (used to depend on the number of taxa and chromosomes) and the homology endpoints (regardless of the number of actual homologies). This is essentially all done automatically in the Compara API

2) Memory leaks. I've simply called each endpoint 1000 times and found a number of places where the gene-trees were not released (they are circular dependencies in this part of the Compara API)

3) Automatic population of the Compara's FullCache adaptors in the Registry preload() method

Other changes:
1) I have moved some code to the Compara API, so that it can be used outside of the REST context.
2) I have added a "subtree_node_id" parameter to the gene-tree endpoints. This is needed by a new web view that shows the tree+multiple alignment of a subtree

There are no format changes, so no version upgrade to request

### DIFF
--- a/lib/EnsEMBL/REST/Controller/GeneTree.pm
+++ b/lib/EnsEMBL/REST/Controller/GeneTree.pm
@@ -101,6 +101,7 @@ sub _set_genetree {
 
   my $prune_species     = $c->request->param('prune_species') ? [$c->request->param('prune_species')] : undef;
   my $prune_taxons      = $c->request->param('prune_taxon') ? [$c->request->param('prune_taxon')] : undef;
+  my ($subtree_filter)  = $c->request->param('subtree_node_id');
   my $aligned           = $c->request()->param('aligned') || $c->request()->param('phyloxml_aligned') ? 1 : 0;
   my $sequence          = $c->request()->param('sequence') || $c->request()->param('phyloxml_sequence') || 'protein';
   my $cigar_line        = $c->request()->param('cigar_line') ? 1 : 0;
@@ -110,6 +111,7 @@ sub _set_genetree {
   $gt->preload(
       -PRUNE_SPECIES => $prune_species,
       -PRUNE_TAXA => $prune_taxons,
+      -PRUNE_SUBTREE => $subtree_filter,
   );
 
   if($self->is_content_type($c, $CONTENT_TYPE_REGEX)) {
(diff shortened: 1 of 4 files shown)


nerdstrike:
I think we'll have to take these changes on trust. I can't see anything wrong, but I can't see any of the business logic either.

magaliruffier:
the test suite is failing because your change to the registry alters the info endpoint
for some reason, the preload of the compara adaptors adds a production_name/assembly alias to the alias hash
can you please have a look?

muffato:
Yes, the Compara API adds the aliases "species_name assembly_name" to the Registry. Since this part runs at preload() time, the aliases now appear in the output of info.t
Is it a problem ? I can change the expected output in info.t

PS: These aliases are not new, they've been added since e24

magaliruffier:
what the test is telling us is that the expected output for species/info has changed
this means that there are new aliases for http://rest.ensembl.org/info/species?content-type=text/xml which were not there before, hence modifying behaviour of an existing endpoint
If we are happy with those aliases being there, this needs to be declared
If they should not be part of this endpoint, the model needs to be updated to filter them out

muffato:
The aliases were never visible in the test-suite, and they were visible on the real REST server only if the worker had processed some compara data before.
I don't think it's new data that has to be declared but a bugfix either way (aliases hidden from the endpoint output, or consistently shown)
As you guys want.

magaliruffier:
is that why our two starmen ended up out of sync? that does sound more like a bugfix, if we have inconsistent behaviour like this.
If showing the data all the time means having to load the compara adaptor, I'd rather have those aliases hidden consistently please

muffato:
> is that why our two starmen ended up out of sync? 

I don't know which "out of sync" issue you are talking about

> If showing the data all the time means having to load the compara adaptor, I'd rather have those aliases hidden consistently please

The point of my commit was precisely to preload the data in the compara adaptors to speed up their first Compara calls. It means that if a core db defines by coincidence "species assembly" as an alias, it will be discarded by the endpoint

nerdstrike:
We had an issue with the REST server reporting different results between requests. The two starmen were not on the same commit somehow.

muffato:
The "species assembly" alias is now discarded, and the test is happy as it is  (no need to change the test data)
ld.t fails for me on farm3, so it may fail here as well. In all the failures it returns [] instead of some data

muffato:
Funnily, ld.t is not complaining on Travis !
I think the errors are related to the funcgen API, but another pair of eyes would be welcome

magaliruffier:
regulation have changed their schema massively for this release
I'll ask them to update the test suite to reflect that, but in the mean time the compara changes are good to go

PR #130, opened 2016-07-01
Support for refseq and merged caches

willmclaren:


magaliruffier:
Looks relatively harmless to me
the test suite is failing due to regulation schema changes, so we can ignore this for the moment
do we need to add any changes to our config files (https://github.com/Ensembl/ensembl-rest_private) ?

willmclaren:
Yes; you'll need to add a line for each of the refseq_dir and merged_dir params, pointing to the equivalent directory to the dir param (but obviously for where you put the refseq or merged cache).

I can put in a PR for what I think this should look like?

PR #132, opened 2016-07-07
Update for Regulation Endpoint for e85 schema changes

juettemann:


magaliruffier:
regulatory.t is still failing
something about expecting an array but returning a hash?

docs
```diff
@@ -1 +1 @@
-Subproject commit de2d85b401801e3bee15e26471f19caeb022bc2d
+Subproject commit 2897209e28cce0658a639862ead23790a3f20ee4
```
magaliruffier: I don't think you should be modifying this subproject, maybe a merge conflict?

magaliruffier:
your merge is not completely clean, we should not see Anja's commits in here
Can you please make sure you rebase against master?

PR #135, opened 2016-07-07
Testing Travis 3 - please ignore.

dzerbino:


magaliruffier:
Almost there!
You can try adding a -verbose option to the runtest command (in the harness.sh file) to get more output
See https://github.com/Ensembl/ensembl/blob/release/84/travisci/harness.sh#L20 for example

magaliruffier:
This is the error:
# GET /eqtl/id/homo_sapiens/ENSG00000223972?content-type=application/json;statistic=p-value;tissue=arm;variant_name=rs4951859
# 2016/07/08 08:06:06 (333) ReturnError.pm 73> Stable ID 'ENSG00000223972' not present in the current Ensembl release 
# Response code 400
not ok 1 - Return p-value for known gene and variant

#   Failed test 'Return p-value for known gene and variant'

PR #138, opened 2016-07-13
fix for annotation variant set

sarahhunt:
Creates the default ensembl version variant set needed by variant annotations


nerdstrike:
Thanks Sarah

lairdm:
Could you remove the "undef" in the return please? You should never return undef, just "return" if there's no value. Thanks.

sarahhunt:
Thanks @lairdm - that's updated.

PR #140, opened 2016-07-18
At7/ld variation

at7:
This came up late reported by a user. We need to filter out variants that are located on non reference sequence.


andrewyatz:
Commenting on this as @lairdm asked me to view it. The change looks okay to me since we're filtering out anything that lives on a patch. That should be reasonably low volume.

:+1:

lairdm:
Looks good, the failing travis is due to an upstream issue in the Bio-HTS tests with ENA's CRAM retrieval being offline.

PR #141, opened 2016-07-20
More compara tests

muffato:
Yeah !
They now cover all the compara endpoints


muffato:
On farm3 the test-suite failed in some variation endpoints. I think it's partly because I didn't have the C thing installed. I hope I haven't messed up any of their endpoints

muffato:
Ah, the test dbs are on the e85 schema but Travis is using master which already has a schema changes for e86. I need to patch the db

muffato:
Funnily, patch_test_database.pl adds "DEFAULT CHARSET=latin1" to every table :)

muffato:
info.t will fail because it expects the compara db to be in the same schema as the core API

magaliruffier:
can you please rebase your fork against master to ensure we have the latest hstlib fixes in?

muffato:
I can see that you've already run the schema_patcher. I don't need 0511efc any more

magaliruffier:
I think that's good to go

t/gene_tree.t
```diff
@@ -0,0 +1,558 @@
+# Copyright [2016] EMBL-European Bioinformatics Institute
+# Copyright [2016] EMBL-European Bioinformatics Institute
```
nerdstrike: Mangled copyright

muffato:
Thank you !

PR #142, opened 2016-07-20
fix docs

sarahhunt:


andrewyatz:
Travis really isn't liking us atmo. The rate limit code was the one that went funny. Your commit seems fine

:+1: though.

lairdm:
Thanks for the cleanup, Sarah.

lairdm:
Oh... umm, could you change pull to be against master please, so we're only pulling updates one direction. Thanks.

PR #144, opened 2016-07-25
remove ExAC plugin docs, add refseq and merged opts

willmclaren:


nerdstrike:
Does this need to be dropped onto the live servers too? We're forced to do a server restart with some regulation data, so we can plop this in too.

willmclaren:
Yeah can't do any harm

lairdm:
Looks reasonable.

PR #145, opened 2016-07-25
add VEP HGVS POST endpoint

willmclaren:
For 86, just an easy thing for me to tick off my list today :-)


willmclaren:
Ah sorry I have messed up the .conf file.

Please don't merge til I've fixed it.

nerdstrike:
We're putting you on hold until 37 is done to keep things clean.

willmclaren:
Just added some more commits not relevant to this PR in particular, but should fix some potential problems with updates to ensembl-variation. That is when everything else is working.

Basically this will make the world a better place.

magaliruffier:
all those lovely dependencies...
At least it's fixed now and it should fix it for the other runs as well
Thanks for the changes

PR #146, opened 2016-07-26
fix tests

sarahhunt:
remove hardcoded release version numbers from tests


nerdstrike:
Smart, I like it.

magaliruffier:
Thanks, that's great :-)
Locally, the tests are still failing for me due to differences in the created and updated string
# |  16|        updated => [                        |        updated => [                        |
# *  17|          '1209464261'                      |          '1209467861'                      *
# |  18|        ],                                  |        ],                                  |

do you know what could be causing this?

sarahhunt:
That's odd - that's the $transcript->modified_date() for ENST00000314040.1 which uses from_date_to_seconds. The value in the test matches what I get from other time converters. Is there something interesting in your local db set up?

magaliruffier:
I'll have to investigate that, but that should affect the PR

PR #147, opened 2016-07-29
neaten variantAnnotation docs

sarahhunt:
reduce column lengths


lairdm:
That should make it wrap better. Thanks.

magaliruffier:
can you please rebase your branch against master to ensure we have the latest hstlib fixes in?

PR #148, opened 2016-08-17
runtime loading for eQTL dependencies

magaliruffier:
with this change, it is possible to run a local REST server without any of the eQTL libraries or data files
with this, we should be able to merge back into master


lairdm:
At the least the require should be in an eval block so it fails more gracefully than taking out the entire starman worker.

magaliruffier:
thanks for the suggestion, I've changed that now
I've also merged with master to include the changes to htslib

nerdstrike:
I think is good to go. I'm not sure if it's there already from Thomas' work, but a big warning at start-up when the EQTL endpoint isn't working would be helpful.

magaliruffier:
the latest change means that the server will start-up even if the endpoint is not working, but not take the whole server down
The endpoints themselves are still listed, but the user gets a 'no EQTL adaptor available' error message
what kind of warning were you thinking of?

nerdstrike:
Ideally something that appears at the console or in the log when you start the server

magaliruffier:
do we want to check the presence of the adaptor, or of eqtl files, or of both?

nerdstrike:
I would go with just the dependency. File discovery and use is more in the hands of the eQTL library itself.

magaliruffier:
I've added a check on all the PERL5LIB libraries
I'm not sure we can do much more without gutting Catalyst again

PR #149, opened 2016-08-19
Should fix failing test

mn1:
Can't call method
"is_current" on an undefined value at
/nfs/users/nfs_m/mr6/src/ensembl-funcgen/modules/Bio/EnsEMBL/
Funcgen/DBSQL/RegulatoryFeatureAdaptor.pm line 950.


magaliruffier:
that's great news, thank you :-)

PR #150, opened 2016-08-23
GA4GH updates for 86

sarahhunt:
Updates to handle the features compliance data in GFF.

Requires https://github.com/Ensembl/ensembl-rest_private/pull/9


lairdm:
There's some GA tests failing, could you take a peek at those please?

sarahhunt:
Good point - I forget them - thanks!

lib/EnsEMBL/REST/Model/ga4gh/ga4gh_utils.pm
```diff
@@ -23,8 +23,9 @@ package EnsEMBL::REST::Model::ga4gh::ga4gh_utils;
 
 use Moose;
 extends 'Catalyst::Model';
-
+use Data::Dumper;
```
lairdm: We probably want to pull Data::Dumper out before putting it in to production

lairdm:
Other than the Data::Dumper, looks sane :)

lib/EnsEMBL/REST/Model/ga4gh/ga4gh_utils.pm
sarahhunt: Agreed!

PR #151, opened 2016-08-24
Feature/eqtl

ilavidas:
patch funcgen test db for e86


magaliruffier:
can you please submit against feature/eqtl_runtime_only instead?
it's got some more recent changes so should allow to test suite to run

PR #152, opened 2016-09-01
Added 0 to frequency, count and allele_count to treat them as numeric…

premanand17:
… and to avoid getting returned as quoted text in results


lairdm:
Makes sense. Yay loosely typed language.

nerdstrike:
I suppose this ought to ideally be done further up the line, rather than us promoting the type in our context.

nerdstrike:
We can just merge this and the travis error will go away on its own.

lairdm:
Yup, Sarah has fixed the underlying issue

PR #153, opened 2016-09-01
update test for overlap variation source

sarahhunt:
VariationFeature summary as hash has been updated to supply source. This updates the test.


nerdstrike:
Fair enough

lairdm:
Thanks for the prompt fix.

PR #154, opened 2016-09-05
Docs change - filter GA4GH variant calls returned

sarahhunt:
Use 3 samples to filter 1KG genotypes for quicker rendering.

Requires: https://github.com/Ensembl/ensembl-rest_private/pull/11


nerdstrike:
Ok

PR #156, opened 2016-09-13
Experimental/jscleanup

andrewyatz:
An attempt to make sure we're using the psgix assets correctly and not loading unnecessary CSS. Also can we just have a quick check this really works. I did this quite a bit ago and TBH I think it was in a working state.


andrewyatz:
The only thing that failed in travis was the rate limiting code. So this has passed (unsurprisingly) but it's a bit annoying rate limit keeps on dying.

```
t/ratelimit.t ............... 18/? 
#   Failed test 'Checking we are limited'
#   at t/ratelimit.t line 313.
#          got: 200
#     expected: 429

#   Failed test 'Checking output is set to the custom message'
#   at t/ratelimit.t line 316.
#          got: 'hello world'
#     expected: 'You have done too much!'
t/ratelimit.t ............... 59/? # Looks like you failed 2 tests of 64.
```

magaliruffier:
Can you rebase against master please?
it should work fine now

magaliruffier:
@andrewyatz can you please rebase against master please?
we should be good to merge in after that

andrewyatz:
Because things had already been done on the flag and pushed here I was somewhat worried about forcing a rebase and force push. I've basically merged again

magaliruffier:
Looks good to go now 👍

nerdstrike:
Has this been tried on a live server?

nerdstrike:
The changes seem to remove syntax highlighting from the code samples. It's an acceptable loss, but I'm not sure if that's what you intended.

magaliruffier:
@andrewyatz any thoughts on Kieron's comment?

andrewyatz:
The intention was to remove the old prettify code and make sure we reference the highlight package. In fact I've not removed them at all

https://github.com/Ensembl/ensembl-rest/pull/156/files#diff-51ae67a92f63c074762e6cd04555ee69R47

However this is *very* old code. The initial commit is over 2 years old. At that time it was working (tested on a running near live server). It is possible your code has progressed on since then and has meant that this code is now breaking. 

TBH I'm not going to put much more effort into this. If you don't want the code then please close this and do not merge.

PR #157, opened 2016-09-20
Feature/eqtl clean

magaliruffier:
with only the commits we want and none of the duplication


nerdstrike:
Pesky rate limit test. I think is go for release.

lairdm:
Pushing the button.

PR #161, opened 2016-09-28
fix: force lookup on correct object type

Zhicheng-Liu:
In Genomes there are objects with different types but the same stable id.
When we try to search for such object by stable id the first record is
picked. That means sometimes we have an object with correct stable id but
wrong type.

This is also affecting Ensembl too:
http://rest.ensembl.org/map/translation/F53H8.4/100..300?content-type=application/json

Make sure that we lookup records with correct object type accroding to endpoints.

Fixes ENSEMBL-4667.


magaliruffier:
good call 👍

lairdm:
:+1: for @nerdstrike's test case request

magaliruffier:
@Zhicheng-Liu is it possible to add a test case for this as suggested?

Zhicheng-Liu:
@magaliruffier Yes it is possible although not straightforward. I was working on it but gave priority to release tasks/bugs half way. I will come back with a test case later.

magaliruffier:
ok, thanks, we'll keep the PR on hold then

nerdstrike:
Excellent, this is ready to go in (minus merge conflicts)

magaliruffier:
nice test cases, thank you
can you try merging your branch with the latest master and see if you can resolve the conflict?

PR #162, opened 2016-10-04
Hotfix/compara/86

muffato:
2 changes after testing test.rest, especially one to allow fetching the mouse-strains trees
I can make the same pull-request against master as well

### DIFF
--- a/lib/EnsEMBL/REST/Model/Lookup.pm
+++ b/lib/EnsEMBL/REST/Model/Lookup.pm
@@ -107,8 +107,9 @@ sub find_genetree_by_member_id {
   }
   Catalyst::Exception->throw("Could not fetch a $object_type object for ID $id") unless $member;
 
+  my $clusterset_id = $c->request->parameters->{clusterset_id};
   my $gta = $dba->get_GeneTreeAdaptor;
-  my $gt = $gta->fetch_default_for_Member($member);
+  my $gt = $gta->fetch_default_for_Member($member, $clusterset_id);
   Catalyst::Exception->throw("No GeneTree found for $object_type ID $id") unless $gt;
   return $gt;
 }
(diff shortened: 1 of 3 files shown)


muffato:
Just a sec, I'll push a unit-test as well

nerdstrike:
Down to the wire eh? Approved.

lairdm:
Waiting for the unit test before these "second eyes" hits merge. Also, @muffato, yes, could you do a pull of this against master too.

lib/EnsEMBL/REST/Model/Lookup.pm
```diff
@@ -148,9 +149,8 @@ sub find_compara_species_sets {
   my $mlsss = $compara_dba->get_MethodLinkSpeciesSetAdaptor->fetch_all_by_method_link_type($method);
   foreach my $mlss (@$mlsss) {
     my $species_set = {};
-    my $species_set_obj = $mlss->species_set_obj();
-    my @species_set_genomes = map { $_->name } @{$species_set_obj->genome_dbs};
-    $species_set->{species_set_group} = $species_set_obj->name;
+    my @species_set_genomes = map { $_->name } @{$mlss->species_set->genome_dbs};
```
magaliruffier: would it be worth keeping the same structure as before and create a species_set variable that is then used in both calls?

muffato:
I've updated the commits, but github doesn't update this page ... Usually it does automatically

magaliruffier:
Have you pushed it? Your branch does not show them either: https://github.com/muffato/ensembl-rest/commits/hotfix/compara/86

muffato:
Yes the commits were there (I've amended them).
I've managed to force a refresh by changing the base branch and back

PR #165, opened 2016-11-01
 Encode error text at HTML so it doesn't integrate into the DOM

brucebolt:
Currently an error in the endpoint's response will integrate into the page's DOM (creating this odd looking header section).  This commit encodes the response as HTML elements, so it appears in the example output box as HTML code.
screen shot 2016-11-01 at 13 42 23


nerdstrike:
Good idea Bruce, I always wondered how that happened. Approved.

PR #167, opened 2016-11-02
Enscoresw 1481

premanand17:


magaliruffier:
one day I'll understand why this rate_limit test case fails randomly
otherwise, looks fine

nerdstrike:
Virtualised system clocks can be less consistent. It depends on the VM platform as well as the OS. This request looks ready to go in.

PR #168, opened 2016-11-11
Enscoresw 1964

premanand17:
Updates to include strain and strain_collection info in REST endpoint info/species - ENSCORESW-1964

### DIFF
--- a/lib/EnsEMBL/REST/Model/Registry.pm
+++ b/lib/EnsEMBL/REST/Model/Registry.pm
@@ -306,14 +306,10 @@ sub get_species {
   my $info = $self->_species_info();
   #have to force numerification again. It got lost ... somewhere
   #filter by division (e.g.: ensembl)
-  if($division){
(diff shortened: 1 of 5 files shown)


lib/EnsEMBL/REST/Model/Registry.pm
```diff
@@ -301,11 +302,30 @@ sub get_species_and_object_type {
...
+    my $new_info_hash = {};
+      foreach my $key(keys $inf){
+        $new_info_hash->{$key} = $inf->{$key} unless $key ~~ @hidden_keys;
+      }
+    push ($new_info_list, $new_info_hash);
+    }
+    return $new_info_list;
+  }
+
+  return [ grep { $_ > 0 } map {$_->{release} += 0; $_} @{$info} ];
```
nerdstrike: Sufficiently obscure to recommend a comment here.

PR #169, opened 2016-11-18
Initial phenotype endpoints

sarahhunt:
These allow extraction of association data by ontology term or accession.


nerdstrike:
You want this in before branch don't you?

nerdstrike:
Changing the test data seems to have inverted some ordered elements in lookup.t. You can leave that to us to fix.

Your code looks ready to go

lairdm:
Seems reasonable.

PR #173, opened 2016-11-25
add code to return warnings as errors when no data comes back

willmclaren:


lib/EnsEMBL/REST/Controller/VEP.pm
```diff
@@ -247,7 +247,15 @@ sub get_consequences {
   my $runner = Bio::EnsEMBL::VEP::Runner->new($config);
   $runner->registry($c->model('Registry')->_registry());
   $runner->{plugins} = $config->{plugins};
+
+  # handle warnings  
+  my $warnings;
```
nerdstrike: This is fun. I think the canonical way to gather errors/warnings from a complex library is to have an error() method within the package that you call after you collect the return values. DBI and libXML both take this approach, and it avoids muddying the scope of packages

We could also dress up the globbing with IO::String

willmclaren: Sounds eminently sensible. Will take a look at doing this.

nerdstrike:
Yeah, that's a lot more pleasing to the eye. Thanks Will

PR #176, opened 2016-12-09
avoid setting plugins key to undef

willmclaren:


lairdm:
We'd rather not add it twice, I'm closing this one and we'll pull it over from master to the branch.

PR #177, opened 2016-12-12
Escape space characters in the URLs for example requests

Zhicheng-Liu:
For some endpoints species alias that contains space charactter could
appear in the URI path. In some languages and tools, URLs are not
automatically encoded. The space characters in the URLs cause failure
in the requests.

This change escapes space characters in the URLs for the example requests
for Curl, Java, Perl and Ruby, according to this document:
http://template-toolkit.org/docs/manual/Filters.html#section_url

Fixes ENSEMBL-4809.


magaliruffier:
Seems sensible

PR #179, opened 2017-01-04
Path to HAL files

muffato:
Hi guys, and happy new year !

During e87 @lairdm told me that we're not configuring the path to access HAL files is not great at all (it's a euphemism). I think it's hard-coded at the moment. Instead, the path should be configured in the main configuration file, and the endpoint should tell the API "This is where you'll find the data".

Here I'm adding a parameter to to the configuration file and pass it to the Compara API (there is a matching change on our side).
I'd still like to keep the same directory structure as on the web server, which is /DATA/PATH/multi/hal_files/*.hal (/DATA/PATH is the bit that's changed in ensembl_rest.conf). Would that fit you ?
If that's OK, then you would just have to rsync Web's /DATA/PATH/multi/ with yours and all the file-based data would work on both sides.

Matthieu

PS: Web directory structure 
```
├── anolis_carolinensis
│   └── AnoCar2.0
│       └── rnaseq
├── astyanax_mexicanus
│   └── AstMex102
│       └── rnaseq
...
├── homo_sapiens
│   ├── GRCh37
│   │   ├── compara
│   │   ├── dna_methylation_feature
│   │   ├── result_feature
│   │   ├── result_set
│   │   ├── rnaseq
│   │   ├── segmentation_file
│   │   └── variation_genotype
│   └── GRCh38
│       ├── compara
│       ├── dna_methylation_feature
│       ├── external_feature_file
│       ├── result_feature
│       ├── result_set
│       ├── rnaseq
│       ├── segmentation_file
│       └── variation_genotype
...
├── multi
│   └── hal_files (only contains murinae_genbank_835.hal at the moment)
...
```

### DIFF
--- a/ensembl_rest.conf.default
+++ b/ensembl_rest.conf.default
@@ -220,6 +220,8 @@ jsonp=1
 <Controller::GenomicAlignment>
   # Sets the default compara to use when finding the best compara fails. Defaults to multi
   # default_compara=multi
+  # Sets the base location of data-files (usually contains multi/hal_files/*.hal)
+  # compara_base_dir_location = /path/to/data/
 </Controller::GenomicAlignment>
 
 <Controller::Variaton>
--- a/lib/EnsEMBL/REST/Controller/GenomicAlignment.pm
+++ b/lib/EnsEMBL/REST/Controller/GenomicAlignment.pm
@@ -26,6 +26,7 @@ require EnsEMBL::REST;
 BEGIN { extends 'Catalyst::Controller::REST'; }
 
 has 'default_compara' => ( is => 'ro', isa => 'Str', default => 'multi' );
+has 'compara_base_dir_location' => ( is => 'ro', isa => 'Str' );
 has 'max_slice_length' => ( isa => 'Num', is => 'ro', default => 1e6);
 
 __PACKAGE__->config(
@@ -56,6 +57,8 @@ sub get_adaptors :Private {
     my $asa = $compara_dba->get_AlignSliceAdaptor();
     my $gata = $compara_dba->get_GenomicAlignTreeAdaptor();
     my $gaba = $compara_dba->get_GenomicAlignBlockAdaptor();
+
+    $mlssa->base_dir_location($self->compara_base_dir_location());
     
     $c->stash(
       compara_dba => $compara_dba,


lairdm:
One of the big differences in rest is we don't have the same swath of files as web. We don't rsync that large directory structure, and we'd rather group the files by functional sections of rest. Particularly with the plan to separate rest out in the multiple instances divided by functional components/endpoints.

For example our structure is:

/data/shared/<release>/<endpoint group>/
where endpoint_group = vep_cache, ga_data...

And in vep_cache, which that path is passed to vep in the config file, Will knows the layout of his directory, and it's however he wishes. He builds the <species>_<grouping> names on the fly himself (ie. homo_sapiens_refseq). Also at deployment time he gives us a tarball that we simply untar in that directory and it lays things out as needed for vep. We tell him via config where his directory is, he manages the layout and poking around in it himself. We'd like to maintain that model, and especially keep configuration details out of code.

For example, we recently went from /mnt/data to /mnt/shared for the data directory root, because we were running out of space in the former. If we had to dig around modules to change all that, that would be bad. Similarly we don't want to have a global data directory everyone uses, for example eqtl data is under /mnt_eqtl_data, a totally different space, again because of space limitations on the rest servers. So being able to tell each functional division of rest independently where its data lives is important.

The two big changes coming with the rest rewrite are auto deployment/configuration, so the configuration file will be built on the fly. And when we move to ebi hardware we're contemplating asking data creators to deploy their data for a release in the appropriate location themselves similar to how groups deposit data for the production team to be rsynced. ie. we'd say release 88 will be ebi:/data/nfs/ensembl/rest/88/ and you'd know you use "compara" under that, you'd put your files there however your modules expect them to be laid out, and on release day the config is built saying "compara = /real-release-dir/rest/88/compara", the directory is rsynced and the service goes live.

So if you want to maintain this multi/, that's fine, but it will likely be under something like /mnt/shared/<release>/compara/, we'd pass in the config directory that as the compara root, and we'd hope to get a tarball that when untared in that directory creates the directory structure the compara endpoints expect.

Is that clearer?

Thanks.

muffato:
Thank you @lairdm . Yes that is clearer. I like the plan whereby each group generates their own section of the data directory and we assume the responsibility of maintaining its structure and syncing it with the API / endpoint configuration options.
We may not want to keep "multi" hanging around, especially because it seems redundant with the "compara" you want to introduce. "multi/hal_files" was a way to pretend supporting multiple divisions (alongside "fungi", "plants", etc, -Ensembl is "multi" in the Registry-) and multiple file-types but currently the mouse-strains HAL file is the only file there :D

You mentioned the idea of splitting the REST server by component. Would this happen within each division ? Or would all the divisions be merged ?

I'll think about that on the way home. Regarding the new configuration option itself, it was really simple to implement in the end, but does that look ok to you ? I know that I'd have to make PRs to ensembl-rest_private as well

Thank you,
Matthieu

lairdm:
This was part of the discussion at the retreat rest session :)

The resource needs for the different endpoints are increasing and differ, the configurations are becoming more complex, and as we discovered with a bug from another team just before Christmas the endpoints can negatively effect each other in some situations. As well, as the data grows we might want to use different technology for different chunks of data or views (ie. tark, mongoose, advanced search, etc). So we have a rough plan to potentially break up the functional groups across different VMS.

As in, you might have a group of VEP rest servers that have a lot of cores, another chunk of compara ones that have lots of memory to deal with the HAL file needs, a group of VMs that do purely sequence lookups. This way we can isolate functional groups and spin up resources as needed for those different divisions. This is all still in the discussion phase as we go through the rest requirements gathering over the next few months.

muffato:
Good to know, thank you !

muffato:
So I think this is fine for now. I have rebased the commit and we'll update ensembl-rest-private later

nerdstrike:
We can work with this, right @lairdm ?

lairdm:
I like, merging. Thanks.

muffato:
Super, thank you !

PR #180, opened 2017-01-11
At7/ld87

at7:
Make population name a required parameter for LD/id and LD/region endpoints.

### DIFF
--- a/lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
+++ b/lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
@@ -72,8 +72,8 @@ sub fetch_LDFeatureContainer_variation_name {
   try {
     $ldfc = $ldfca->fetch_by_VariationFeature($vf, $population);
   } catch {
-    Catalyst::Exception->throw("Something went wrong during LD computation") if ($_ =~ /TEST LD calc errror/);
-    Catalyst::Exception->throw("Something went wrong while fetching from LDFeatureContainerAdaptor") if ($_ =~ /TEST LD calc errror/);
+    $c->log->error("LD endpoint for $variation_name $population_name $window_size caused an error: $_");
+    Catalyst::Exception->throw("LD computation for $variation_name $population_name $window_size caused an error.");
   };
   return $self->to_array($ldfc)
 }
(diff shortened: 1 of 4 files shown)


lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
```diff
@@ -64,27 +63,23 @@ sub fetch_LDFeatureContainer_variation_name {
...
-    }
-    my $ldfc;
-    try {
-      $ldfc = $ldfca->fetch_by_VariationFeature($vf, $population);
-    } catch {
-      Catalyst::Exception->throw("Something went wrong while fetching from LDFeatureContainerAdaptor");
-    };
-    return $self->to_array($ldfc)
+  my $pa = $c->model('Registry')->get_adaptor($species, 'Variation', 'Population');     
+  my $population = $pa->fetch_by_name($population_name);
```
nerdstrike: Can this throw Ensembl errors?

lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
```diff
@@ -99,23 +94,18 @@ sub fetch_LDFeatureContainer_slice {
...
+  if (!$population) {
+    Catalyst::Exception->throw("Could not fetch population object for population name: $population_name");
   }
-  my $ldfc = $ldfca->fetch_by_Slice($slice);
-  return $self->to_array($ldfc);
+  my $ldfc;
+  try {
+    $ldfc = $ldfca->fetch_by_Slice($slice, $population);
+  } catch {
+    Catalyst::Exception->throw("Something went wrong while fetching from LDFeatureContainerAdaptor");
```
nerdstrike: Does the ldfca error make it to the log files, or have you swallowed it here on purpose?

lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
at7: The PopulationAdaptor returns undef if the population isn't stored in our database.
population_name is now required for this endpoint. I already check in the controller if population_name is given and throw an exception if not.

lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
at7: How do I write to the log files?

nerdstrike: I presume you can find the logger through the context - $c->log->error($_)

nerdstrike: Otherwise we just "something went wrong" and we can't debug it.

at7: I'll add that. thanks

nerdstrike:
I think this is ok for integration now

PR #183, opened 2017-01-18
At7/ld widget

at7:
This functionality has been requested by the web team. It is needed for the LD widget and reduces the number of required rest calls. With mu updates to ld_vcf.c it should be fast enough.

### DIFF
--- a/root/documentation/ld.conf
+++ b/root/documentation/ld.conf
@@ -1,7 +1,7 @@
 <endpoints>
   <ld_id_get>
     description=Computes and returns LD values between the given variant and all other variants in a window centered around the given variant. The window size is set to 500 kb.    
-    endpoint=ld/:species/:id/:population_name/:d_prime/:r2/:window_size/:attribs
+    endpoint=ld/:species/:id/:population_name
     method=GET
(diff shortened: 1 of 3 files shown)


root/documentation/ld.conf
```diff
@@ -1,7 +1,7 @@
 <endpoints>
   <ld_id_get>
     description=Computes and returns LD values between the given variant and all other variants in a window centered around the given variant. The window size is set to 500 kb.    
-    endpoint=ld/:species/:id/:population_name/:d_prime/:r2/:window_size
+    endpoint=ld/:species/:id/:population_name/:d_prime/:r2/:window_size/:attribs
```
nerdstrike: I believe this isn't correct. :variable implies part of the URL, not simply ?options you can slap on the end

at7: I see. Only the required parameters are listed here? then I also need to remove :d_prime/:r2/:window_size. Is that correct?

nerdstrike: That's more like it. endpoint is a URL spec, the options are enumerated below

magaliruffier:
I think this is not good for merging

PR #184, opened 2017-02-01
At7/gwas/ld update

at7:
Reduce output for endpoint ld/id with attribs. Has been asked for by Ridwan for GWAS LD widget


magaliruffier:
these fields are all new to this release, is that correct?
in other words, these changes are not on the current live endpoint?

at7:
Correct. And they will only appear if used with attribs=1. This is foremost for the GWAS LD widget

avullo:
Trying to understand here, does the reduced output preserve the amount of information returned?

at7:
the endpoint receives an input variant and the computes pairwise LD stats for the input variant and all other variants in that region. because you know already everything about your input variant we only return attribs for the second variant. The use case is so far only the LD widget which joins the results of two endpoints. In order to avoid another REST call we put consequence and location information in the ld/id endpoint

avullo:
Sounds sensible, thanks for the explanation.

PR #185, opened 2017-02-03
At7/gwas ld clinsgn

at7:
Another small addition. We would like to add clinical significance to the z menu on  the LD widget. So here it is in the output for the ld/id endpoint.


avullo:
That's fine, only thing is you're testing with empty content, would it be better to add some?

magaliruffier:
Hi @at7 can you please look at updating the tests with some values?

PR #186, opened 2017-02-20
Fix/var tests

willmclaren:


lib/EnsEMBL/REST/Model/Variation.pm
```diff
@@ -192,7 +192,7 @@ sub get_gene_phenotype_info {
 
   foreach my $phen (@pfs) {
     my $hash = $self->phen_as_hash($phen);
-    my $key = join("", sort values %$hash);
+    my $key = join("", sort grep {$_} values %$hash);
```
nerdstrike: This will need a comment, or someone will take it out in future since it appears to do nothing.

willmclaren: Why would you say it appears to do nothing?

nerdstrike: I am putting myself of future coder who goes "huh?".

willmclaren: grep {$_} means your checking it has a non-0 and non-null value. How would you assume that does nothing unless you don't know what grep does?

nerdstrike: If you understand all Perl primitives, you understand all code right? No comments about circumstance required at all QED. Commenting on at least the purpose of the sub{} will save someone some brain glucose in future.

PR #187, opened 2017-02-23
patch funcgen test dbs

ilavidas:
Apply patch 'patch_87_88_c.sql' to funcgen test db


magaliruffier:
thanks for the patch, it looks fine

PR #188, opened 2017-02-24
fix: get strain info from separate query

Zhicheng-Liu:
For most (if not all) of species found in EG's collection databases, "species.strain"
and "species.strain_collection" are not available as the meta_key in meta tables.
Adding these two with others will return no matching rows.

Fix ENSEMBL-4870.

### DIFF
--- a/lib/EnsEMBL/REST/Model/Registry.pm
+++ b/lib/EnsEMBL/REST/Model/Registry.pm
@@ -376,28 +376,32 @@ sub _build_species_info {
         else {
           $dbc->sql_helper->execute_no_return(
             -SQL => q/select m1.meta_value, m2.meta_value, m3.meta_value, m4.meta_value
-		from meta m1, meta m2, meta m3, meta m4, meta m5, meta m6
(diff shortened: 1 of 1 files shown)


Zhicheng-Liu:
It seems to me the test failures were caused by the incompatible software version (88) and database version (87).

avullo:
Have you tried running the test suite locally with the expected correct matching api/db versions?

Zhicheng-Liu:
Good point! I will do that later and report back.

Zhicheng-Liu:
@avullo, I have run the test suite from my local checkout. The failed test cases are the same for with or without my changes. So I believe my changes will not cause issues.

avullo:
But have you set the rest server configuration to point to 87 DBs and, correspondingly checked out the 87 branch?

Zhicheng-Liu:
Yes. That was the case.

avullo:
Ok, let me try to replicate the issue

avullo:
Your changes can go, apart from that we need to make the testing suite more flexible wrt the particular release we want it to run on.

PR #189, opened 2017-02-24
fix: get strain info from separate query

Zhicheng-Liu:
For most (if not all) of species found in EG's collection databases, "species.strain"
and "species.strain_collection" are not available as the meta_key in meta tables.
Adding these two with others will return no matching rows.

Fix ENSEMBL-4870.


magaliruffier:
looks sensible enough, I'm afraid we are not always exhaustive when testing for multi-species dbs
do you need the change in release/87 branch as well or could it wait for next release?

Zhicheng-Liu:
Yes, we need the change in release/87 as this issue comes from an RT ticket. I have another pull request for the same change to the release/87 here: https://github.com/Ensembl/ensembl-rest/pull/188

PR #191, opened 2017-02-28
Feature/vep add flags

willmclaren:


magaliruffier:
special code for ucsc, just what I like ;-)
looks fine to me

willmclaren:
Yeah, that's in support of a feature I haven't actually implemented, but might in future. Would just be a config change to enable it.

lib/EnsEMBL/REST/Controller/VEP.pm
```diff
@@ -45,7 +41,9 @@ our $DOWNSTREAM_DISTANCE_BAK;
 # /vep/:species
 sub get_species : Chained('/') PathPart('vep') CaptureArgs(1) {
   my ( $self, $c, $species ) = @_;
-  $c->stash->{species} = $c->model('Registry')->get_alias($species);
+  my $reg = $c->model('Registry');
+  $c->stash->{species} = $reg->get_alias($species);
+  $c->stash->{assembly} = $reg->get_adaptor($species, 'core', 'coordsystem')->fetch_all->[0]->version;
```
lairdm: Err, can we please do some checking and not blindly do ->[0] ?

lairdm:
ok, I'm placated now.

PR #194, opened 2017-03-16
fixed example's input errors in documentation file

waakanni:
This was causing the examples on my test rest server to failed. The input parameters were clearly wrong so all the issues with the last pull request must have led to an untested version of the documentation file being pushed.  

All appear to be fixed now.


nerdstrike:
That particular error isn't your fault. It's a chain of dependencies falling over.

magaliruffier:
can you submit the fix against release/88 as well please?

PR #199, opened 2017-04-18
Use placeholder in phenotype region endpoint documentation

Zhicheng-Liu:
Related to ENSEMBL-4883.


magaliruffier:
travis failure is unrelated, looks good to me

lairdm:
yay non-hardcoded identifiers!

PR #204, opened 2017-05-12
Beacon

helensch:
Endpoint for GA4GH Beacon

A pull request will be made GA4GH_beacon conf values.

Response include JSON boolean values and URLs  (is escaping of "/"  needed for JSON  response).


avullo:
Hi Helen, this looks fine, tests are present and pass, will merge if nobody else objects.

Thanks

PR #206, opened 2017-05-25
Changed display_id to stable_id

juettemann:
Removing deprecated message from server log

### DIFF
--- a/lib/EnsEMBL/REST/Model/Regulatory.pm
+++ b/lib/EnsEMBL/REST/Model/Regulatory.pm
@@ -146,7 +146,7 @@ sub get_probe_info {
     my $transcripts = [];
     for my $tm (@$transript_mappings) {
       my $hash = {};
-      $hash->{display_id}   = $tm->display_id();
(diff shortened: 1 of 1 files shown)


nerdstrike:
Is this fixing the right thing? The deprecation was on display_id(), not the hash key.

For anyone else looking, this was the warning:

------------------ DEPRECATED ---------------------
Deprecated method call in file /home/ubuntu/src/ensembl-rest/lib/EnsEMBL/REST/Model/Regulatory.pm line 149.
Method Bio::EnsEMBL::Funcgen::ProbeTranscriptMapping::display_id is deprecated.
display_id has been deprecated and will be removed in Ensembl release 92. Please use stable_id instead.
Ensembl API version = 89
---------------------------------------------------

nerdstrike:
Apologies for the **very bold deprecation message** . Thanks for submitting the change.

juettemann:
Thanks for spotting, changed now the right content. Doing it aside the rest course, through checks appreciated. ;)

PR #211, opened 2017-06-22
update haplotypes tests to reflect change in variation API

willmclaren:


magaliruffier:
Thanks for ensuring Travis and Core stay happy :-)

PR #212, opened 2017-07-03
Endpoint to retrieve variant records by publication

helensch:
2 configuration options to be added, variation_pmid, variation_pmcid


nerdstrike:
Looks like we're due a patch to our variation test DB, but your code looks fine.

magaliruffier:
latest failure will be fixed with Will's PR, this change is good to go

PR #213, opened 2017-07-27
UpDownDistance VEP plugin replaced by built in --distance flag

willmclaren:


willmclaren:
The remaining test failure looks like a funcgen thing, certainly wouldn't be because of anything I've changed - @mn1 @ilavidas any ideas?

nerdstrike:
Looks like we need to patch our test DB with regulation schema changes. The rate limit one may go away on its own. Anything time-limited on Travis is a bit inconsistent.

avullo:
Will, can you pull the latest changes on master? We've just patched the test DBs to integrate fungen and other schema changes. The overlap endpoint test now pass.

PR #214, opened 2017-07-31
Tj1/probeset endpoint

juettemann:
New endpoint for probeset

### DIFF
--- a/bin/production/start_rest_server.sh
+++ b/bin/production/start_rest_server.sh
@@ -36,7 +36,8 @@ fi
 APP_HOME=$(cd $SCRIPT_DIR/../../ && pwd)
 
 # Library path work
-for ensdir in ensembl-variation ensembl-funcgen ensembl-compara ensembl ensembl-io ensembl-hdf5; do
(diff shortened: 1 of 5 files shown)


avullo:
Thanks Thomas.
We've patched the test DBs on master, including human funcgen, and a previously merged PR now makes the trascript haplotype test succeed. Unfortunately, this conflicts with your version of test funcgen schema.
Instead of resolving the conflict manually, would it be ok if we close this PR so that you open another one from the current master with:
- the probeset endpoint code/documentation;
- TESTS FOR THE PROBESET ENDPOINT (this is required if we want to accept);
- no commit about the startup script, if it's not making any changes.

juettemann:
Sure, no problem. Could you advise how I get the startup script out of the commit?
Thanks!

avullo:
I was more in favor of creating another PR on a different branch with no startup commit included. About removing commits, I don't if it's possible, but if it is my feeling is that it's complicated ;)

magaliruffier:
It is easier to copy commits to a new branch than removing commits from an existing branch
Something like this should do the trick:
git checkout master
git pull
git branch feature/probeset_endpoint
git cherry-pick b3f944f198f85abc24
git cherry-pick 1e22de89433
git commit
git push

PR #215, opened 2017-08-02
add VariantRecoder endpoint

willmclaren:


lib/EnsEMBL/REST/Controller/VariantRecoder.pm
```diff
@@ -0,0 +1,147 @@
...
+with 'EnsEMBL::REST::Role::PostLimiter';
+
+# /variant_recoder/:species
+sub species : Chained('/') PathPart('variant_recoder') CaptureArgs(1) {
+  my ( $self, $c, $species ) = @_;
+  my $reg = $c->model('Registry');
+  $c->stash->{species} = $reg->get_alias($species);
+
+  my $csa = $reg->get_adaptor($species, 'core', 'coordsystem');
+  $c->go('ReturnError', 'from_ensembl', ['Unable to fetch CoordSystem adaptor']) unless $csa;
```
magaliruffier: the 'from_ensembl' type ReturnError implies there might be a throw in the initial call. If that's the case, it would be better to have a try/catch block, otherwise we'll just get a 500 without any error message

lib/EnsEMBL/REST/Controller/VariantRecoder.pm
```diff
@@ -0,0 +1,147 @@
...
+  my $css = $csa->fetch_all;
+  $c->go('ReturnError', 'from_ensembl', ['No coordinate systems found']) unless $css && @$css;
```
magaliruffier: same here for the ReturnError

lib/EnsEMBL/REST/Controller/VariantRecoder.pm
```diff
@@ -0,0 +1,147 @@
...
+  my $user_config = $c->request->parameters;
+  my $config = $self->_include_user_params($c,$user_config);
+
+  $self->get_results($c, $config, $rs_id);
+}
+
+sub id_POST {
+  my ($self, $c) = @_;
+
+  if (!$c->stash->{has_variation}) { $c->go('ReturnError', 'custom', ["Species ".$c->stash->{species}." does not have a variation database"]); }
```
magaliruffier: this can probably go in the sub id, as it is shared by both the GET and POST methods

t/variation.t
```diff
@@ -29,6 +29,7 @@ use Test::Differences;
 use Catalyst::Test ();
 use Bio::EnsEMBL::Test::MultiTestDB;
 use Bio::EnsEMBL::Test::TestUtils;
+use Storable qw(dclone);
```
magaliruffier: is Storable already part of the normal packages? otherwise it should be added to the cpanfile as well

t/variation.t
```diff
@@ -82,4 +83,122 @@ my $expected_result = { rs142276873 => $expected_variation_1, rs67521280 => $exp
...
+delete($expf2->[0]->{$_}) for qw(hgvsc hgvsp);
+
+is_json_POST(
+  $base,
+  '{"ids" : ["rs200625439", "rs142663151"], "fields": "hgvsg,id"}',
+  [$expf->[0], $expf2->[0]],
+  'variant_recoder - POST - restrict fields'
+);
+
+
```
magaliruffier: any chance you could add some error tests? testing for expected behaviour when not providing a species or a dummy one for example

t/variation.t
willmclaren: Storable is a core module

lib/EnsEMBL/REST/Controller/VariantRecoder.pm
willmclaren: This code is copied from the VEP controller which you guys wrote...

magaliruffier: And it is barfing badly there: http://rest.ensembl.org/vep/random/id/COSM476?content-type=application/json

willmclaren: Can you suggest a fix then we can apply it to both

magaliruffier:
```
-  my $csa = $reg->get_adaptor($species, 'core', 'coordsystem');
-  $c->go('ReturnError', 'from_ensembl', ['Unable to fetch CoordSystem adaptor']) unless $csa;
-  my $css = $csa->fetch_all;
-  $c->go('ReturnError', 'from_ensembl', ['No coordinate systems found']) unless $css && @$css;
+  my ($csa, $css);
+  try {
+    $csa = $reg->get_adaptor($species, 'core', 'coordsystem');
+  } catch {
+    $c->go('ReturnError', 'from_ensembl', [qq{$_}]) if $_ =~ /STACK/;
+    $c->go('ReturnError', 'custom', [qq{$_}]);
+  };
+  try {
+    $css = $csa->fetch_all;
+  } catch {
+    $c->go('ReturnError', 'from_ensembl', [qq{$_}]) if $_ =~ /STACK/;
+    $c->go('ReturnError', 'custom', [qq{$_}]);
+  };
```

I've got it ready for the VEP controller, will push once the test databases are updated

t/variation.t
magaliruffier: Perfect, thanks 👍

lib/EnsEMBL/REST/Controller/VariantRecoder.pm
willmclaren: removed, this check was not needed

t/variation.t
willmclaren: error tests now added

lib/EnsEMBL/REST/Controller/VariantRecoder.pm
willmclaren: Thanks, added

magaliruffier:
Looking good, thanks

PR #217, opened 2017-08-08
The name of the alignment has changed

muffato:


nerdstrike:
Ok. Not sure what's up with VEP.

avullo:
Fake failure, and it doesn't depend on the here proposed changes, good to go.

PR #218, opened 2017-08-30
ENSCORESW-2287

jimmyebi:


lib/EnsEMBL/REST/Model/BiotypeGroup.pm
```diff
@@ -0,0 +1,129 @@
...
+
+sub _validate {
+  my ($self, $c, $inputParameter, $existingValues) = @_;
+
+  my %parameters = map { $_ => 1 } @{$existingValues};
+  if(exists($parameters{$inputParameter})) {
+
+#  returns true in Perl...
+   return 1;
+  }else{
```
lairdm: spacing. but if that's the biggest criticism I can come up with, everything else must be aces!

lairdm:
I know it's a bigger refactor than we'd probably be interested in, but I'm not sure how I feel about ensembl-production and ensembl-orm becoming dependencies of REST. ensembl-production just doesn't feel like it should be public facing like that, rather than for production tasks. And ensembl-orm, so we have our orm within the api, plus this other orm. Eek. Feels like we're adding to the maintenance burden.

But this might be a bigger discussion than this one pull request.

.travis.yml
```diff
@@ -26,6 +26,7 @@ before_install:
     - git clone --branch 1.3.2 --depth 1 https://github.com/samtools/htslib.git
     - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-production.git
     - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-orm.git
+    - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl.git
```
lairdm: That is already clones about 8 lines up.

PR #219, opened 2017-09-01
set default reference allele for insertion

sarahhunt:
To resolve "Missing reference or alternate sequence" issue


magaliruffier:
I can confirm the fix is working on the REST server
Unfortunately, Travis is not happy with the new default linux distro and I'd like to get that fixed before we pull the change in
The change is on one GRCh37 server now, so at least the example is working again

magaliruffier:
Ok, Travis has been fixed, if you can just rebase your PR, it should pick up the change

sarahhunt:
@magaliruffier - that looks better!

PR #220, opened 2017-09-04
set default reference allele for insertion

sarahhunt:
Also update master


nerdstrike:
Is this trivially testable? Regression testing and all that.

magaliruffier:
@sarahhunt is there a way of testing this simply or should be pull this in as it is?

avullo:
@sarahhunt sorry to bother again, can you please reply to Mag?

thanks

lairdm:
I was hoping to branch today or Monday, are we pulling this in before the branch? Are we able to add a test? Thanks.

sarahhunt:
@lairdm - after e!90 was released, a problem was reported and this change was quickly added to branch to fix it. It will need to go into master. I had hoped to find the time to add a test but have not done so.

lairdm:
Fair enough. I'll merge it now so I can branch, but scout's honour a test is forthcoming before e92? :)

PR #222, opened 2017-10-30
Add searching compara dbs for a gene id at homology endpoint

Zhicheng-Liu:
In EnsemblGenomes we have vertebrate gene ids stored in compara pan homology
database. These ids are not loaded into stable_id_lookup database. When a user
wants to fetch homologues through a vertebrate gene id, it returns error because
the id is not found in stable_id_lookup database.

This change allows searching compara dbs at homology endpoint so that when a
vertebrate gene id is not found in the stable_id_lookup database, we continue to
search for it in compara dbs, so that the id could be used to fetch homologues
through EnsemblGenomes's rest endpoint.

Related to ENSEMBL-5061.

### DIFF
--- a/lib/EnsEMBL/REST/Controller/Homology.pm
+++ b/lib/EnsEMBL/REST/Controller/Homology.pm
@@ -50,7 +50,9 @@ sub get_adaptors :Private {
 
   try {
     my $species = $c->stash()->{species};
-    my $compara_dba = $c->model('Registry')->get_best_compara_DBAdaptor($species, $c->request()->param('compara'), $self->default_compara());
+    my $param_compara = $c->request()->param('compara');
+    my $default_compara = $self->default_compara();
+    my $compara_dba = $c->model('Registry')->get_best_compara_DBAdaptor($species, $param_compara, $default_compara);
     my $gma = $compara_dba->get_GeneMemberAdaptor();
     my $ha = $compara_dba->get_HomologyAdaptor();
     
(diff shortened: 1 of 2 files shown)


lib/EnsEMBL/REST/Controller/Homology.pm
```diff
@@ -50,7 +50,9 @@ sub get_adaptors :Private {
 
   try {
     my $species = $c->stash()->{species};
-    my $compara_dba = $c->model('Registry')->get_best_compara_DBAdaptor($species, $c->request()->param('compara'), $self->default_compara());
+    my $param_compara = $c->request()->param('compara');
```
avullo: This change does not seem to add anything useful, except some improved readability.

avullo:
I reckon this is not trivial to test, but I guess you've done that somehow?!

lib/EnsEMBL/REST/Controller/Homology.pm
Zhicheng-Liu: No, it is more than that. In fact it fixes a bug.
The arguments for a subroutine call in Perl are passed as a list. The second argument to get_best_compara_DBAdaptor here is $c->request()->param('compara') which will be evaluated in a LIST context.
The problem is that in LIST context, the Catalyst::Request returns an empty list for an argument that does not exist:
https://github.com/perl-catalyst/catalyst-runtime/blob/master/lib/Catalyst/Request.pm#L783
So when the request url does not have 'compara=' parameter, the arguments list passed to get_best_compara_DBAdaptor here is [$species, (), $default_compara] which is [$species, $default_compara], causing second argument to be skipped.
The change here is to evaluate the Catalyst::Request::param call in a SCALAR context so that it returns 'undef' for a parameter that does not exist.

Zhicheng-Liu:
I have not written any test cases to back up this change.  But I have test this change with EnsemblGenomes REST as discussed in ENSEMBL-5061.

lib/EnsEMBL/REST/Controller/Homology.pm
avullo: Thank you for the explanation, we've learned something more about Catalyst :+1:, at least me.
Might have been good to have it explained at PR creation time, thanks anyway.

avullo:
Ok, tests in production mode are equally, if not more, indicative of matched expected behaviour.

PR #224, opened 2017-11-02
Test the three different headers for passing the remote user's ip add…

lairdm:
…ress in to the client:

REMOTE_ADDR
REMOTE_USER
HTTP_X_CLUSTER_CLIENT_IP

Not the cleanest method, using a global variable, but forcing the header name in to the pseudo-middleware used to hack the headers doesn't work without deriving fully from the Middleware class.

### DIFF
--- a/t/ratelimit.t
+++ b/t/ratelimit.t
@@ -30,12 +30,13 @@ use Plack::Middleware::EnsThrottle::Second;
 use Plack::Middleware::EnsThrottle::Minute;
 use Plack::Middleware::EnsThrottle::Hour;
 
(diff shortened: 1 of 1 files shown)


avullo:
This could certainly go for me, but when I saw it I didn't have a clue as to where it was coming from and why. I believe it's still our team practice to start the commit message with the JIRA ticket it, whenever there's one, e.g. "[ENSEMBL-5102]. Test the three different headers for passing the remote user's ip ...". At least this is what I always do, always nice to have a reference.

avullo:
Why has this been closed?

PR #227, opened 2017-11-08
Variation/ld updates

at7:
Minor updates. 1) deal with warnings generated by LDFeatureContainerAdaptor in variation API. 2) Define population_name first before it is used for writing log entries.


avullo:
Brilliant, this is good to go for me, and thanks for the quick variation fix!

avullo:
ratelimit failed, this should be just random noise.

lairdm:
Merging before branching today.

PR #233, opened 2018-01-05
Master

aparton:
### Requirements

### Description

_One or more sentences describing in detail the proposed changes._

Added in a transcript_id flag to allow for filtering by transcript

### Use case

Requested - allows for filtering by transcript on the REST side rather than having to be performed by the user

### Benefits

_If applicable, describe the advantages the changes will have._

### Possible Drawbacks

Flag is only available on rest - slight discord between REST and CL VEP

### Testing

Yes.

Pass, of course.

Yes.

### Changelog

Added the ability to filter VEP output by transcript

### DIFF
--- a/lib/EnsEMBL/REST/Controller/VEP.pm
+++ b/lib/EnsEMBL/REST/Controller/VEP.pm
@@ -28,7 +28,7 @@ use Try::Tiny;
 require EnsEMBL::REST;
 
 EnsEMBL::REST->turn_on_config_serialisers(__PACKAGE__);
-BEGIN {
+BEGIN { 
   extends 'Catalyst::Controller::REST';
 }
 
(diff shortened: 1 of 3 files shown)


lib/EnsEMBL/REST/Controller/VEP.pm
```diff
@@ -353,6 +356,17 @@ sub _include_user_params {
   return \%vep_params;
 }
 
+sub _amend_transcript_id {
+  my ($self, $config) = @_;
+
+  if(exists($config->{transcript_id}) && length($config->{transcript_id}) < 25 && not ($config->{transcript_id} =~ m/[^A-Z0-9_.]/i))
```
lairdm: This is perl... we use ! ;)
Nah, we've discussed this in Core, just for consistency, perhaps change not to ! as you're using && just before rather than and. And while you're at it, please wrap that line, it's longer than the github viewer, again for readability in the long term.
Nitpicky, but a simple change. Thanks.

lairdm:
Looks good except that one formatting quibble. Approved once that's change.

PR #234, opened 2018-01-09
Anja 92/sv overlap endpoint

at7:
### Description
This is a bug fix: The overlap endpoint for structural_variation never worked the way it was implemented. There was a change for the somatic_structural_variation routine. I updated the structural_variation routine accordingly.
There are SO terms for the variant consequence and for the variant class. For structural variants we have only SO terms for the variant class. We could consider to have a filter for variant class in the future. But this is beyond the scope of this bug fix.
### Use case
get all structural variants or all somatic structural variants in a region
### Benefits
bug fix
### Possible Drawbacks
none
### Testing
I provided 2 more tests to ensure that structural_variation and somatic_structural_variation return data.


lib/EnsEMBL/REST/Model/Overlap.pm
```diff
@@ -345,13 +345,13 @@ sub variation {
 
 sub structural_variation {
   my ($self, $slice) = @_;
-  my @so_terms = $self->_get_SO_terms();
-  my ($source, $include_evidence, $somatic) = (undef)x3;
-  my $sv_class = (@so_terms) ? $so_terms[0] : ();
+#  my @so_terms = $self->_get_SO_terms();
```
lairdm: If you're removing code could you please remove and not comment it out. We can always go back through the history to find it again. Thanks.

lairdm:
Looks good, thanks.

PR #236, opened 2018-01-19
The /info/comparas endpoint returns the version of the compara database, not the core database

muffato:
### Description and Use case

The /info/comparas test assumes that the test compara database is in the same version as the test core database. This assumption is not always met (e.g. before branching the API, or when developing locally).
The commit solves the issue by comparing the version returned by the endpoint to the one of the database itself

### Benefits

No more Travis failures

### Possible Drawbacks

None

### Testing

Yes.

Pass, of course.

Yes.

### DIFF
--- a/t/info.t
+++ b/t/info.t
@@ -42,17 +42,20 @@ Catalyst::Test->import('EnsEMBL::REST');
 require EnsEMBL::REST;
 
 my $VERSION = software_version();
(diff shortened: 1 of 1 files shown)


avullo:
A perfect follow up of the discussion we've had. This is good for me, thanks @muffato :+1:

PR #237, opened 2018-01-24
added new REST endpoint /phenotype/gene/:species/:gene

ima23:
### Requirements

- corresponding  ensembl-rest_private pull request

### Description

Added a new REST endpoint  /phenotype/gene/:species/:gene with two options: include_associated and include_overlap.
It works both with gene Ensembl stable identifiers and gene symbols. 

### Use case

/phenotype/gene/homo_sapiens/MLXIPL
/phenotype/gene/homo_sapiens/SAP30BP?include_associated=1
/phenotype/gene/homo_sapiens/SAP30BP?content-type=application/json;include_associated=1;include_overlap=1

### Testing

added tests: t/phenotype_gene.t


lib/EnsEMBL/REST/Controller/Phenotype.pm
```diff
@@ -116,6 +116,39 @@ sub region: Chained('/') PathPart('phenotype/region') Args(2) ActionClass('REST'
...
+
+
+application/json
+
+=cut
+sub gene_GET {}
+
+sub gene: Chained('/') PathPart('phenotype/gene') Args(2) ActionClass('REST') {
+  my ($self, $c, $species, $gene) = @_;
+
```
nerdstrike: I'd like to see some primitive argument checking here, i.e. that both $species and $gene are required arguments.

lib/EnsEMBL/REST/Model/Phenotype.pm
```diff
@@ -22,6 +22,7 @@ package EnsEMBL::REST::Model::Phenotype;
 use Moose;
 use Catalyst::Exception qw(throw);
 use Scalar::Util qw/weaken/;
+use List::MoreUtils qw(uniq);
```
nerdstrike: This is a new dependency for ensembl-rest. Please add it to the /cpanfile

lib/EnsEMBL/REST/Model/Phenotype.pm
```diff
@@ -199,4 +200,73 @@ sub fetch_features_by_region {
...
+      $record->{attributes}->{risk_allele}           = $pf->risk_allele()           if $pf->risk_allele();
+      $record->{attributes}->{clinical_significance} = $pf->clinical_significance() if $pf->clinical_significance();
+      $record->{attributes}->{external_reference}    = $pf->external_reference()    if $pf->external_reference(); ## from study - eg OMIM
+      $record->{attributes}->{external_id}           = $pf->external_id()           if $pf->external_id();        ## from attrib eg ClinVar
+      $record->{attributes}->{p_value}               = $pf->p_value()               if $pf->p_value();
+      $record->{attributes}->{odds_ratio}            = $pf->odds_ratio()            if $pf->odds_ratio();
+      $record->{attributes}->{beta_coefficient}      = $pf->beta_coefficient()      if $pf->beta_coefficient();
+      $record->{attributes}->{associated_gene}       = $pf->associated_gene()       if $pf->associated_gene();
+
+      if ($record) {
```
nerdstrike: Useless if

lib/EnsEMBL/REST/Model/Phenotype.pm
```diff
@@ -199,4 +200,73 @@ sub fetch_features_by_region {
...
+      my $gene_specific_slice = $slice_ad->fetch_by_gene_stable_id($g->stable_id());
+      my @pfs_overlap = @{$phenfeat_ad->fetch_all_by_Slice($gene_specific_slice)};
+      push @pfs, @pfs_overlap;
+    }
+    @pfs = uniq @pfs;
+
+    while (my $pf = shift @pfs){
+      my $ontology_accessions = $pf->get_all_ontology_accessions();
+
+      my $record =  { description       => $pf->phenotype_description(),
```
nerdstrike: I'm not convinced this is precisely the right place for this. The composition of a "phenotype feature" object for serialisation is not necessarily bound to the functioning of our REST API, and could be handled by the PhenotypeFeature itself, or else be placed in lib/EnsEMBL/REST/EnsemblModel . This would help us with future remodelling of our REST service architecture.

t/phenotype_gene.t
```diff
@@ -0,0 +1,190 @@
...
+my $gene_name='FOXF2'; 
+
+
+#Null based queries
+{
+  my $gene_name = 'gene123H';
+  my $msg = "Gene gene123H not found.";
+  action_bad_regex("$base/$gene_name", qr/gene123H/, $msg);
+}
+
```
nerdstrike: Perhaps test the parameter validation by making a null request to the server with missing arguments.

t/phenotype_gene.t
```diff
@@ -0,0 +1,190 @@
...
+	      external_reference => 'PMID:8626802'  
+	    },
+	    description => 'soft palate development',
+	    location => '6:1312675-1314992',
+	    source => 'GOA' 
+          }
+        ];
+	
+  my $json = json_GET("$base/$gene", 'Gene with 2 phenotype features associated from GOAs ');
+  is(scalar(@{$json}), $expected, 'Gene with 2 phenotype associations found');
```
nerdstrike: Try to resist the urge to use string-matching tests on numeric comparisons. They generally work, but there are pathological exceptions. cmp_ok($got,'==',$expected,$message) is safer.

lib/EnsEMBL/REST/Model/Phenotype.pm
sarahhunt: @nerdstrike - a good point, but this fits with the style of our other endpoints. We want to review all of our data models for the next iteration of REST, so it would be a good time to address it then.

sarahhunt:
re popularity - we've had a couple of requests for this functionality and it is a common use case. I'm not sure how this translates into usage stats.

nerdstrike:
@sarahhunt As long as you have no expectations of it being as successful/server-killing as the LD endpoints, then it'll be fine.

lib/EnsEMBL/REST/Model/Phenotype.pm
nerdstrike: As you wish.

ima23:
@nerdstrike Thanks for the feedback! I've updated the branch. 
The curl times for BRCA1 are: 1.2s for simple URI, 6.652s for include_associated and 6.066s for include_associated and include_overlap together.

nerdstrike:
My concerns are addressed - over to @lairdm

nerdstrike:
Wow, 6 seconds? That's a bit serious if the endpoint gets heavily used.

sarahhunt:
@nerdstrike - BRCA1 is about as bad as it gets. Most genes are instant.

ima23:
The examples for rest.ensembl.org take between 0.142s and 0.858s.

nerdstrike:
@premanand17 I think we should be able to get this merged before branch. Pester @lairdm if you don't wish to review it yourself.

PR #246, opened 2018-04-04
Biotype endpoints

tgrego:
### Requirements

### Description

Added some biotype info endpoints. Namely info/biotypes/groups and info/biotypes/name
Existing info/biotypes remains unchanged.

### Use case

info/biotypes/groups provides the list of available groups.
info/biotypes/groups/:group/:object_type provides the list of biotypes within that group.
info/biotypes/name/:name/:object_type provides the list of biotypes with that name.

### Benefits

Finding biotype data to be used elsewhere.

### Possible Drawbacks

None expected.

### Testing

Yes.

Pass, of course.

Yes.

### Changelog

[info/biotypes/]
- Added new route [info/biotypes/groups] to retrieve list of available biotype groups.
- Added new route [info/biotypes/groups/:group/:object_type] to retrieve list the properties of biotypes within provided group (and object_type).
- Added new route [info/biotypes/name/:name/:object_type] to retrieve list the properties of biotypes with provided name (and object_type).

### DIFF
--- a/lib/EnsEMBL/REST/Controller/Info.pm
+++ b/lib/EnsEMBL/REST/Controller/Info.pm
@@ -192,7 +192,7 @@ sub biotype_group : Path("biotypes/groups") CaptureArgs(2) ActionClass('REST') {
   };
 
   ## Return 404 for get requests with no return
-  if ( !$biotypes->[0] ) {
+  if ( ! scalar @{$biotypes} ) {
     my $ot_error = '';
     if ($object_type) { $ot_error = " and object_type $object_type" }
     $c->go( 'ReturnError', 'not_found', ["biotypes not found for group $group$ot_error"]);
@@ -220,7 +220,7 @@ sub biotype_name : Path("biotypes/name") CaptureArgs(2) ActionClass('REST') {
   };
 
   ## Return 404 for get requests with no return
-  if ( !$biotypes->[0] ) {
+  if ( ! scalar @{$biotypes} ) {
     my $ot_error = '';
     if ($object_type) { $ot_error = " and object_type $object_type" }
     $c->go( 'ReturnError', 'not_found', ["biotypes not found for name $name$ot_error"]);
--- a/lib/EnsEMBL/REST/Controller/Info.pm
+++ b/lib/EnsEMBL/REST/Controller/Info.pm
@@ -189,6 +189,9 @@ sub biotype_group : Path("biotypes/groups") Args(1) ActionClass('REST') {
     $c->go('ReturnError', 'custom', [qq{$_}]);
   };
 
+  ## Return 404 for get requests with no return
+  $c->go( 'ReturnError', 'not_found', ["biotypes not found for group $group"]) unless $biotypes->[0];
+
   $self->status_ok($c, entity => $biotypes);
   return;
 }
@@ -209,6 +212,9 @@ sub biotype_name : Path("biotypes/name") CaptureArgs(1) ActionClass('REST') {
     $c->go('ReturnError', 'from_ensembl', [qq{$_}]) if $_ =~ /STACK/;
     $c->go('ReturnError', 'custom', [qq{$_}]);
   };
+
+  ## Return 404 for get requests with no return
+  $c->go( 'ReturnError', 'not_found', ["biotypes not found for name $name"]) unless $biotypes->[0];
 
   $self->status_ok($c, entity => $biotypes);
   return;
--- a/lib/EnsEMBL/REST/Controller/Info.pm
+++ b/lib/EnsEMBL/REST/Controller/Info.pm
@@ -143,15 +143,83 @@ sub external_dbs_GET :Local :Args(1) {
   return
 }
 
-sub biotypes :Local :ActionClass('REST') :Args(1) { }
+sub biotypes_GET { }
 
-sub biotypes_GET :Local :Args(1) { 
+sub biotypes : Chained('/') PathPart("info/biotypes") Args(1) ActionClass('REST') {
   my ($self, $c, $species) = @_;
   my $dba = $c->model('Registry')->get_DBAdaptor($species, 'core', 1);
   $c->go('ReturnError', 'custom', ["Could not fetch adaptor for species $species"]) unless $dba;
   my $obj_biotypes = EnsEMBL::REST::EnsemblModel::Biotype->get_Biotypes($c, $species);
   my @biotypes = map { $_->summary_as_hash() } @{$obj_biotypes};
   $self->status_ok($c, entity => \@biotypes);
+  return;
+}
+
+sub biotype_groups_GET { }
+
+sub biotype_groups : Chained('/') PathPart("info/biotypes/groups") Args(0) ActionClass('REST') {
+  my ($self, $c) = @_;
+
+  my $groups;
+
+  try {
+    $groups = $c->model('Biotype')->fetch_biotype_groups;
+
+  } catch {
+    $c->go('ReturnError', 'from_ensembl', [qq{$_}]) if $_ =~ /STACK/;
+    $c->go('ReturnError', 'custom', [qq{$_}]);
+  };
+
+  $self->status_ok($c, entity => $groups);
+  return;
+}
+
+
+sub biotype_group_GET { }
+
+sub biotype_group : Chained('/') PathPart("info/biotypes/groups") Args(1) ActionClass('REST') {
+  my ($self, $c, $group) = @_;
+
+  my $biotypes;
+
+  try {
+    $biotypes = $c->model('Biotype')->fetch_biotypes_by_group($group);
+
+  } catch {
+    $c->go('ReturnError', 'from_ensembl', [qq{$_}]) if $_ =~ /STACK/;
+    $c->go('ReturnError', 'custom', [qq{$_}]);
+  };
+
+  $self->status_ok($c, entity => $biotypes);
+  return;
+}
+
+sub biotype_names_GET { }
+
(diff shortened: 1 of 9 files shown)


lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -155,6 +155,74 @@ sub biotypes_GET :Local :Args(1) {
...
+    $c->go('ReturnError', 'custom', [qq{$_}]);
+  };
+
+  $self->status_ok($c, entity => $biotypes);
+  return;
+}
+
+sub biotype_names_GET { }
+
+sub biotype_names : Chained('/') PathPart("info/biotypes/name") Args(0) ActionClass('REST') {
```
nerdstrike: Do I understand correctly that you've made an entire route for detecting a missing parameter?

lib/EnsEMBL/REST/Model/Biotype.pm
```diff
@@ -0,0 +1,142 @@
...
+Retrieves list of biotype groups
+
+=cut
+
+sub fetch_biotype_groups {
+  my ($self) = @_;
+
+  my $dbc = $self->context->model('Registry')->get_adaptor( 'Homo_sapiens', 'Core', 'Biotype' )->db->dbc();
+
+  my $sql = "SELECT DISTINCT biotype_group FROM biotype WHERE biotype_group IS NOT NULL ORDER BY biotype_group";
```
nerdstrike: Can the Biotype classes not do this? Given the impending branch day, perhaps this goes on the TODO pile.

lib/EnsEMBL/REST/Model/Biotype.pm
```diff
@@ -0,0 +1,142 @@
...
+Retrieves list of biotypes part of a provided biotype group
+
+=cut
+
+sub fetch_biotypes_by_group {
+  my ($self, $group) = @_;
+
+  my $dbc = $self->context->model('Registry')->get_adaptor( 'Homo_sapiens', 'Core', 'Biotype' )->db->dbc();
+
+  my $sql = "SELECT name, biotype_group, object_type, so_acc FROM biotype WHERE biotype_group = \"$group\" ORDER BY object_type, name";
```
nerdstrike: See above. Not a fan of Ensembl schema knowledge in the server model.

t/info.t
```diff
@@ -147,6 +147,36 @@ is_json_GET(
   action_bad_regex('/info/biotypes/wibble', qr/Could not fetch adaptor for species .+/, 'Bogus species means error message');
 }
 
+# /info/biotypes/groups
+{
+  my $biotypes_json = json_GET('/info/biotypes/groups', 'Get the list biotype groups');
+  is(ref($biotypes_json), 'ARRAY', 'Array wanted from endpoint');
+  cmp_ok(scalar(@{$biotypes_json}), '==', 8, 'Ensuring we have the right number of biotype groups');
```
nerdstrike: A weak test. Perhaps you could sort and spot-check the string value of the first or last element?

lib/EnsEMBL/REST/Controller/Info.pm
tgrego: yes...
without that route it goes straight into the old info/biotypes/:species endpoint, complaining the species 'name' does not exist.
I can however make the check there and remove that useless route.

lib/EnsEMBL/REST/Model/Biotype.pm
tgrego: The core BiotypeAdaptor does not do this as of now.
Can easily be added as a method in the BiotypeAdaptor and used by the rest api.
was however a quick way to see it working without code dependencies...

t/info.t
tgrego: indeed, will do that

lib/EnsEMBL/REST/Model/Biotype.pm
nerdstrike: I suggest you create a new issue on JIRA, tag it for helpdesk and leave it for another time.

tgrego: created ENSCORESW-2695 and tagged it for helpdesk.

lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -148,6 +148,9 @@ sub biotypes_GET { }
 sub biotypes : Chained('/') PathPart("info/biotypes") Args(1) ActionClass('REST') {
   my ($self, $c, $species) = @_;
   my $dba = $c->model('Registry')->get_DBAdaptor($species, 'core', 1);
+  if ( $species eq 'name') {
```
nerdstrike: How does this work?

magaliruffier:
Two comments:
- The response blob contains details about which object_type a biotype applies to, but it would also be nice to be able to filter on it, so I don't even get distracted by all those transcript biotypes I am not interested in
- The database type is available in the biotype table but not on the endpoint. It would be good to be able to filter by that, as most people are interested in biotypes from the core database, ie the main gene set. In particular, there are some RefSeq specific biotypes in the otherfeatures database, from the RefSeq import which do not exist in Gencode and I can imagine the confusion from users if they see we have guide_RNA as a biotype, but can never find any in the Ensembl gene set.
We should probably return by default only data from the core database and allow advanced users to query for special databases if they want, similar to what we are doing with the lookup endpoint.

root/documentation/info.conf
```diff
@@ -165,7 +165,57 @@
...
+      <two>
+        path=/info/biotypes/groups/
+        capture=__VAR(biotype_group)__
+        content=application/json
+      </two>
+    </examples>
+  </biotypes_groups>
+
+  <biotypes_name>
+    description=List the biotypes for a given biotype name. Usually two exist for a particular name, based on the object_type (gene or transcript).
```
nerdstrike: I would like the wording here to be changed to something like "List the properties of biotypes with a given name".

lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -155,6 +155,65 @@ sub biotypes_GET :Local :Args(1) {
...
+
+sub biotype_name : Path("biotypes/name") CaptureArgs(1) ActionClass('REST') {
+  my ($self, $c, $name) = @_;
+
+  $c->go('ReturnError', 'custom', ["Missing argument ':name' for endpoint info/biotypes/name/:name"]) unless $name;
+
+  my $biotypes;
+
+  try {
+    $biotypes = $c->model('Biotype')->fetch_biotypes_by_name($name);
```
nerdstrike: This silently fails when a request is given for a biotype that doesn't exist. An empty listref is the response. Normally it would be more conventional to get a 404 with an explanatory message.

root/documentation/info.conf
```diff
@@ -165,7 +165,57 @@
...
     </examples>
   </biotypes>
-  
+
+  <biotypes_groups>
+    description=List the functional classifications within a biotype group. Without argument the list of available biotype groups is returned.
+    endpoint="info/biotypes/groups/:group"
+    method=GET
+    group=Information
+    output=json
```
nerdstrike: Technically your endpoints seem to support text/html (YAML-like), and text/xml

lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -189,6 +189,9 @@ sub biotype_group : Path("biotypes/groups") Args(1) ActionClass('REST') {
     $c->go('ReturnError', 'custom', [qq{$_}]);
   };
 
+  ## Return 404 for get requests with no return
+  $c->go( 'ReturnError', 'not_found', ["biotypes not found for group $group"]) unless $biotypes->[0];
```
nerdstrike: Does this produce unitialised warnings when it triggers? It would probably be less warny to check the list size.

lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -176,45 +176,55 @@ sub biotype_groups : Path("biotypes/groups") Args(0) ActionClass('REST') {
...
+    $biotypes = $c->model('Biotype')->fetch_biotypes_by_group($group, $object_type);
 
   } catch {
     $c->go('ReturnError', 'from_ensembl', [qq{$_}]) if $_ =~ /STACK/;
     $c->go('ReturnError', 'custom', [qq{$_}]);
   };
 
   ## Return 404 for get requests with no return
-  $c->go( 'ReturnError', 'not_found', ["biotypes not found for group $group"]) unless $biotypes->[0];
+  if ( !$biotypes->[0] ) {
```
nerdstrike: A similar observation here. I always get nervous around direct element access without checking it exists first.

lib/EnsEMBL/REST/Model/Biotype.pm
```diff
@@ -81,11 +81,13 @@ Retrieves list of biotypes part of a provided biotype group
...
 
 sub fetch_biotypes_by_group {
-  my ($self, $group) = @_;
+  my ($self, $group, $object_type) = @_;
 
   my $dbc = $self->context->model('Registry')->get_adaptor( 'Homo_sapiens', 'Core', 'Biotype' )->db->dbc();
 
-  my $sql = "SELECT name, biotype_group, object_type, so_acc FROM biotype WHERE biotype_group = \"$group\" ORDER BY object_type, name";
+  my $ot_filter = '';
+  if ( $object_type ) { $ot_filter = "AND object_type = \"$object_type\"" }
```
nerdstrike: In future you can use qq() to negate teh need to escape quotes.

PR #249, opened 2018-04-12
update maf to af

at7:
### Description
We updated the return keys for the vep endpoints from *_maf to *_af. The change appears in the colocated variants section of the output hash.

### Use case
Users want to know if their input variant has know frequency data. We report the frequency from co-located variants the input allele matches one of the alleles of the colocated variant. We then add the frequency to the output. MAF stands for minor allele frequency. This is not always the case when we return the frequency. It is better to call it AF which just

### Benefits
Avoids confusion and sometimes naming errors. The allele frequency is not always the minor allele frequency.
We made the change in the VEP script version and VEP tool. This would make our reporting more consistent across the VEP applications.

### Possible Drawbacks
It is very likely that users have written parsers which filter for *_maf.

### Testing

vep.t has been updated.

vep.t runs without errors.
- tests fail on travis. not related to the PR's change.

### Changelog

[/vep/:species/*] 
We updated the frequency keys in the collocated variants entry. The key has been updated from *_maf to *_af. The new naming better reflects that we only report allele frequencies from co-located variants if the input allele is matched.


premanand17:
Looks good. Test fails because of t/genomic_alignment.t, which should pass after merging the pull request https://github.com/Ensembl/ensembl-rest/pull/248

PR #252, opened 2018-05-02
export phenotype ontology accessions

ima23:
### Description

The phenotype REST endpoint retrieves the existing phenotypes but currently does not report the associated ontology accession terms. This is now being updated. 

### Use case

### Benefits

The result of the endpoint will now also contain ontology accessions.

### Possible Drawbacks

None.

### Testing

Yes. The corresponding tests have been updated. 

Tests pass with release/93 branch code.

Yes.

### Changelog

The endpoint functions as before, just the output is enriched with accession terms if they exist.

### DIFF
--- a/lib/EnsEMBL/REST/Model/Phenotype.pm
+++ b/lib/EnsEMBL/REST/Model/Phenotype.pm
@@ -221,6 +221,7 @@ sub fetch_features_by_gene {
   unless (defined $gene_ad ) {Catalyst::Exception->throw("Species $species not found.");}
 
   my $phenfeat_ad = $c->model('Registry')->get_adaptor($species,'variation', 'phenotypefeature');
+  $phenfeat_ad->_include_ontology(1);
   my $slice_ad = $c->model('Registry')->get_adaptor($species, "core", "slice");
 
   my $include_assoc = $c->request->parameters->{include_associated};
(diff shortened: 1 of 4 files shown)


lairdm:
Failed tests:
archive - due to branching and master now being release/94, will be fixed with my travis pull branch change
gene_tree, genomic_alignment - Compara have been made aware of the test breaking bcause of changes in their repos, they're investigating. Not relevant to this pull request.
ratelimit - the usual, timing issue

lib/EnsEMBL/REST/Model/Phenotype.pm
```diff
@@ -221,6 +221,7 @@ sub fetch_features_by_gene {
   unless (defined $gene_ad ) {Catalyst::Exception->throw("Species $species not found.");}
 
   my $phenfeat_ad = $c->model('Registry')->get_adaptor($species,'variation', 'phenotypefeature');
+  $phenfeat_ad->_include_ontology(1);
```
nerdstrike: On balance, perhaps _include_ontology() should not have an underscore prefix if it was intended to be called from outside the object. I won't ask you to change it here, but it is unconventional

ima23: Thanks for checking @nerdstrike. Yes, well spotted and totally agree. Will sort this out in one of the future releases, as it involves deeper look into the API.

PR #253, opened 2018-05-02
export phenotype ontology accessions

ima23:
### Description

The phenotype REST endpoint retrieves the existing phenotypes but currently does not report the associated ontology accession terms. This is now being updated.

### Use case

### Benefits

The result of the endpoint will now also contain ontology accessions.

### Possible Drawbacks

None.

### Testing

Yes. The corresponding tests have been updated.

Tests pass.

Yes.

### Changelog

The endpoint functions as before, just the output is enriched with accession terms if they are present.


lairdm:
Parallel to PR 252.

PR #256, opened 2018-05-18
remove use of private method

ima23:
### Description

Removing the private method call to reflect the bug fix in the dependent repository.

### Use case

### Benefits

The result of the endpoint will contain ontology accessions without the need of calling a private method.

### Possible Drawbacks

None.

### Testing

No change in the tests was needed.

Tests pass.

Yes.

### Changelog

The endpoint functionality is not changed.


nerdstrike:
Thanks for refactoring this @ima23 . 

Approve when ready @premanand17

PR #257, opened 2018-05-23
add new privacy policy link on top banner

jyothishnt:
### Description
New GDPR privacy policy link

### Benefits

Users can enjoy reading the updated policy

### Possible Drawbacks

We may lose users if they don't accept the privacy policy

### Testing

Tested successfully in sandbox


premanand17:
Merged as the test failures unrelated to the current changes.

PR #264, opened 2018-07-03
Feature/ld endpoint

magaliruffier:
This is not a formal pull request
This feature branch contains changes to the LD endpoint that have not been integrated into master.
This should be reviewed to see if the changes should be integrated or the branch deleted.


at7:
Don't merge the changes into master and delete the branch. I have a local copy of the changes and if we need to add the code in the future I can commit a fresh PR. The code was required for the LD widget. But I haven't heard anything from the webteam about the requirements.

magaliruffier:
Thanks for the update, I'll delete the branch

PR #265, opened 2018-07-03
Tutorial/stats endpoint

magaliruffier:
This is not a formal pull request
The changes in this branch are not part of master, they should be reviewed and integrated or discarded


lairdm:
I'm ok deleting it. We had talked about wanting a way to disable endpoints, which this has functionality for. If two others agree it offers benefit, we can merge it. Otherwise if I don't hear back I'll delete the branch next week.

avullo:
I second the integration: 1. provides an endpoint which might be useful to develop from sometimes in the future (web could benefit while working on their alpha release site; 2. the disabling endpoint functionality is something which makes the current REST a little closer to what it's supposed to look like in the future.

PR #270, opened 2018-07-03
Add new privacy policy link for archive 88 site

imran-ebi:
### Description

New GDPR privacy policy link for archive 88 site.

### Benefits

Users can access the updated privacy policy.

### Possible Drawbacks

We may lose users if they don't accept the privacy policy.

### Testing

Tested successfully in sandbox.


avullo:
Good.

PR #273, opened 2018-07-06
add new options to phenotype endpoints and report MIM ids by default

ima23:
### Description

The default result reporting for phenotype endpoints (gene, region, accession, term) has been changed to report MIM ids if present.
New optional parameters (include_review_status, include_pubmed_id and include_submitter) have been added. All three parameters are available for gene and region endpoints, while the accession and term endpoints will have all but the include_submitter option.
Additional code changes where performed to reflect the latest perl API methods.

### Use case

The MIM id has been requested and the new options allow for more details to be retrieved.

### Benefits

The update allows retrieval of MIM ids, PubMed ids, review status and submitter names.

### Possible Drawbacks

None.

### Testing

Yes. The corresponding tests have been updated and new ones added.

Tests pass.

Yes.

### Changelog

The four phenotype endpoints will report by default MIM ids if they are present.
[/phenotype/accession/:species/:accession] Added optional parameters include_review_status and include_pubmed_id
[/phenotype/accession/:species/:term] Added optional parameters include_review_status and include_pubmed_id
[/phenotype/gene/:species/:gene] Added optional parameters include_submitter, include_review_status and include_pubmed_id
[/phenotype/gene/:species/:region] Added optional parameters include_submitter, include_review_status and include_pubmed_id


ima23:
Thanks for checking @nerdstrike! Yes, the aim is to have the code parts simplified in the future. Currently there are small differences between the endpoints results.

lairdm:
My only objection would be the summary_as_hash() rather than the old method using accessors to build the objects. You're making a hash, populating it with things beyond what you're going to need, doing a memcpy of all the elements, then throwing it away to create a new hash, again and again. It just feels a little wasteful and computationally expensive. But not a show stopper, especially if there's a rationalization coming for the copy-pastiness Kieron raised in a future release.

PR #275, opened 2018-07-13
Feature/fixratelimit

nerdstrike:
### Description

The rate limiter has been intermittently failing on Travis for some time now. There is clearly a problem with time on a heavily virtualised platform.

### Use case
If we make the rate limiter test less dependent on time, it should be less fragile in CI environments

### Benefits
Travis failures should be real problems, and not just the rate limiter. ratelimit.t should finish sooner as it has no significant sleep calls to wait for.

### Possible Drawbacks
May not have fixed the rate limiter for all time. There are still some parts which rely on a time component.

### Testing
No changes outside of ratelimit.t

Travis tests have passed 3 times

### Changelog

No user-facing changes

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -65,3 +65,4 @@ test_requires 'Test::Deep';
 test_requires 'HTTP::Request::Common';
 test_requires 'Plack::Test';
 test_requires 'Net::CIDR::Lite';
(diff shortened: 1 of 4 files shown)


premanand17:
Merging to see if it works..

PR #278, opened 2018-07-31
Add endpoint checker script

nicklangridge:
### Description

Selenium test script to check that all endpoint documentation examples load without error.

### Use case

After setting up a REST server, run the script to check that all endpoints appear to be working.

### Benefits

Helps to pick up obviously broken endpoints or missing documentation params. Useful when setting up the staging/test/live servers.

### Possible Drawbacks

Can't think of any.

### Testing

n/a

### Changelog

n/a

### DIFF
--- a/selenium/scripts/check_endpoints.pl
+++ b/selenium/scripts/check_endpoints.pl
@@ -1,4 +1,4 @@
-#!/usr/local/bin/perl
+#!/usr/bin/env perl
 
 =head1 LICENSE
 
--- a/selenium/scripts/check_endpoints.pl
+++ b/selenium/scripts/check_endpoints.pl
@@ -2,14 +2,12 @@
 
 =head1 LICENSE
 
-Copyright [2009-2016] EMBL-European Bioinformatics Institute
-
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2018] EMBL-European Bioinformatics Institute
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
-
-    http://www.apache.org/licenses/LICENSE-2.0
-
+     http://www.apache.org/licenses/LICENSE-2.0
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/selenium/scripts/check_endpoints.pl
+++ b/selenium/scripts/check_endpoints.pl
@@ -1,0 +1,106 @@
+#!/usr/local/bin/perl
+
+=head1 LICENSE
+
+Copyright [2009-2016] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=head1 DESCRIPTION
+
+This script checks all endpoints in the REST server to see if they load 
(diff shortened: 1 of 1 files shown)


selenium/scripts/check_endpoints.pl
```diff
@@ -0,0 +1,106 @@
+#!/usr/local/bin/perl
```
mkszuba: I believe the recommended way of invoking the interpreter these days is "/usr/bin/env perl".

selenium/scripts/check_endpoints.pl
```diff
@@ -0,0 +1,106 @@
...
+done_testing();
+exit 0;
+
+
+sub test_endpoints {
+  ok $selenium->open($base_url), 'homepage';
+  foreach my $href (get_hrefs_like('/documentation/info')) {   
+    ok $selenium->open($href) && eval { page_loads_without_error() }, $href;
+  }
+}
```
mkszuba: Wouldn't hurt to add "return;" here, just in case.

selenium/scripts/check_endpoints.pl
```diff
@@ -0,0 +1,106 @@
...
+
+test_endpoints();
+done_testing();
+exit 0;
+
+
+sub test_endpoints {
+  ok $selenium->open($base_url), 'homepage';
+  foreach my $href (get_hrefs_like('/documentation/info')) {   
+    ok $selenium->open($href) && eval { page_loads_without_error() }, $href;
```
mkszuba: Please add parentheses to all ok() calls so that it is visually unambiguous which arguments are passed where. This one in particular, while not particularly difficult to decode, is already packed enough with content to require a pause to mentally do so.

selenium/scripts/check_endpoints.pl
```diff
@@ -0,0 +1,106 @@
+#!/usr/local/bin/perl
+
+=head1 LICENSE
+
+Copyright [2009-2018] EMBL-European Bioinformatics Institute
```
mkszuba: This should specify joint Sanger/EMBL-EBI copyright for 1999-2015 and EMBL-EBI one for 2018-2018, shouldn't it?

selenium/scripts/check_endpoints.pl
nerdstrike: With plenv it is imperative that we don't accidentally invoke the default system Perl.

selenium/scripts/check_endpoints.pl
nicklangridge: No, my understanding is that EBI-only copyright is correct.

nicklangridge: But I'm not sure what dates should be used - let me check

mkszuba: Hmph. Had a look at _Ensembl Coding Guidelines_ on Confluence, they only mentioned the joint copyright. On the other hand, the _Ensembl Software Licence_ mentions the combined one but without having bumped the year from 2017 to 2018.

I give up.

selenium/scripts/check_endpoints.pl
nicklangridge: done

selenium/scripts/check_endpoints.pl
nicklangridge: done

selenium/scripts/check_endpoints.pl
nicklangridge: seems subjective, but done anyway

selenium/scripts/check_endpoints.pl
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env perl
+
+=head1 LICENSE
+
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2018] EMBL-European Bioinformatics Institute
```
mkszuba: Okay, the Development Coordinator hath spoken: drop the joint line and change the copyright years in the EBI-only one to simply "[2018]".

selenium/scripts/check_endpoints.pl
nicklangridge: Aha. I changed it to match the existing copyright in the REST repo (i.e. with sanger attribution) for consistency after discussing with @andrewyatz.

mkszuba:
Okay then, consider rebasing instead of merging to get rid of the merge commit (we do not seem to have many of them) and as far as I am concerned you should be all set.

nicklangridge:
OK, merge commit removed

PR #279, opened 2018-07-31
ENSCORESW-2814

tgrego:
### Description

Some replacement variables were not set.

### Use case

An error would appear when visiting http://rest.ensembl.org/documentation/info/biotypes_groups 
stating the missing variables.

### Benefits

The error should go away

### Possible Drawbacks

none I can see

### Testing

no

### Changelog

no

### DIFF
--- a/ensembl_rest.conf.default
+++ b/ensembl_rest.conf.default
@@ -181,6 +181,10 @@ jsonp=1
     genomic_alignment_species2=gallus_gallus
     genomic_alignment_pw_region=2:106041430-106041480:1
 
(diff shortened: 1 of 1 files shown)


magaliruffier:
These changes need to be added to ensembl-rest_private, as this is where the configuration is set for the REST server deployment.
We should also update our internal documentation to ensure new endpoints are correctly set

tgrego:
Ah, very well. Then one change will fix it everywhere.
Didn't know where it was, will delete this branch

PR #280, opened 2018-07-31
Fix broken URL pointing to the Ensembl Variation documentation

ens-lgil:
### Description

Fix old URL pointing to the Ensembl Variation documentation

### Benefits

Link is pointing to an actual page on the Ensembl Live Website

### DIFF
--- a/root/documentation/overlap.conf
+++ b/root/documentation/overlap.conf
@@ -52,7 +52,7 @@
       </so_term>
       <variant_set>
         type=String
-        description= Short name of a set to restrict the variants found. (<a target="_blank" href="http://www.ensembl.org/info/genome/variation/data_description.html\#variation_sets">See list of short set names</a>) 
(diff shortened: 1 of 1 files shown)


root/documentation/overlap.conf
```diff
@@ -52,7 +52,7 @@
       </so_term>
       <variant_set>
         type=String
-        description= Short name of a set to restrict the variants found. (<a target="_blank" href="http://www.ensembl.org/info/genome/variation/data_description.html\#variation_sets">See list of short set names</a>) 
+        description= Short name of a set to restrict the variants found. (<a target="_blank" href="//www.ensembl.org/info/genome/variation/species/sets.html">See list of short set names</a>)
```
lairdm: Why was the protocol dropped from the href tag? Would that not break it?

ens-lgil: Normally it should pick up the default protocol (i.e. httpS for Ensembl).

PR #281, opened 2018-08-02
Depend on modules used by built-in YAML and HTML serialisers

mkszuba:
### Description

Add YAML::XS, YAML::Syck and URI::Find to the cpanfile as required modules.

### Use case

These three modules are required for the YAML (the former) and HTML (the
latter two) serialisers provided by Catalyst::Controller::REST to work.
The former is especially important because it is the fallback
serialiser for unknown content types, which in our case includes e.g.
text/x-fasta; its presence is therefore necessary for erroneous requests
for our special content types (such as requesting a sequence for an
unknown transcript stable ID) to return the actual error message, e.g.

	{"error":"ID 'ENSTdeadbeef' not found"}

(the content type of error responses to requests involving content types
handled by the fallback serialiser is 'application/json') instead of the
generic (and pretty much useless)

	Content-Type text/html is not supported.

### Benefits

Prevents the REST server from swallowing errors on special content types. See also: https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-2709 .

### Possible Drawbacks

Slightly increased installation size.

### Testing

N/A


magaliruffier:
Some of these modules are missing from the EBI farm environment, causing tests to fail when run locally
please ensure they get installed in the development environment

PR #283, opened 2018-08-03
Nickl logpid paths

nicklangridge:
### Description

Adding code to create log root and PID root dirs when server starts.
It is expected that log root and PID dirs exist and have mode 755 and currently the server will fail to start if this is not the case. Seems to make sense to create them if needed.

### Use case

This is especially useful when deploying code in the EBI nfs-based environment where the same code checkout will be running on multiple VMs and log/pid dir names are based on the hostnames. For each host new log/pid dirs must be created (currently a manual process).

### Benefits

Simplifies deployment SOPs

### Possible Drawbacks

Can't think of any

### Testing

Only manual testing

### Changelog

n/a

### Target branches

Please can this go to master and release/94 branches


lairdm:
This is a long needed cleanup. @nicklangridge could you do a parallel pull request against release/94 please. Thanks.

lairdm:
@nicklangridge just pinging you on this pull request, if you could address the checking for the existence of the directory before attempting to create it. Thanks.

nicklangridge:
@lairdm re. existence check, the make_path function in https://metacpan.org/pod/File::Path does it's own check so I didn't think I needed to do it here.

avullo:
From the documentation: "If the directory already exists (and thus does not need to be created), the permissions will not be modified."
Hence the only problem I see is in the case of an existing directory with different permissions. If I'd want to be pedantic, I'd check for the existence and mode before invoking make_path.

nicklangridge:
OK, that's fair enough. And do what if the permissions are found to be wrong? give a warning and try to change them? die if they can't be changed?

nicklangridge:
I've added a check for existence and permissions. If a dir cannot be created or permissions cannot be set then the script will die and the server won't start.

I also updated the 'Remove cruft' commit to take out an old/redundant hashbang that I noticed.

If this is ok I'll make another PR for release/94

PR #294, opened 2018-08-30
Create log and PID file paths at server start-up (release/94)

nicklangridge:
See comments here https://github.com/Ensembl/ensembl-rest/pull/283


nicklangridge:
Got this wrong, closing

PR #301, opened 2018-09-24
Add missing perl module dependencies (for release/95)

nicklangridge:
See https://github.com/Ensembl/ensembl-rest/pull/300


mkszuba:
Definitely needed but let's hold off merging until Travis stops failing on missing release branches, just in case.

PR #303, opened 2018-09-26
Bump version to 8.0, e94 - for real this time

mkszuba:
Turns out the commit which was supposed to bump the version, did not. Given there will have been breaking changes between e93 and e94, we'd better make sure the server version number does get updated before the release day.


mkszuba:
Correct. Or to be precise, even the docs were not completely changed - the changelog was updated but the version number in Home.md was not.

PR #304, opened 2018-09-26
E95 API version and docs

mkszuba:
Version bump of the API and documentation in preparation for the branching next week.


mkszuba:
We should be good to go now.

PR #305, opened 2018-09-26
For population allele frequencies only return allele count if available

helensch:
### Requirements

### Description

For population allele frequencies only return the allele count if available

### Use case

Currenty if there is allele_frequency but no allele_count the end_point returns the allele_count as 0.
This is incorrect.  The allele_count key is no longer returned.

### Benefits

The allele_count is not reported if not available.

### Possible Drawbacks

Users may have looked for the 'allele_count' key

### Testing

A test has been added for a population allele frequency with no allele_count

The test passes

Yes

### Changelog

/variation/:species/:id?pops=1,  for population allele frequencies with no allele count property is not returned.

### DIFF
--- a/lib/EnsEMBL/REST/Model/Variation.pm
+++ b/lib/EnsEMBL/REST/Model/Variation.pm
@@ -330,7 +330,7 @@ sub pops_as_hash {
   my $population;
   $population->{frequency} = 0 + $allele->frequency(); # Add 0 to treat it as numeric (to avoid quoting)
   $population->{population} = $allele->population->name();
-  $population->{allele_count} = 0 + $allele->count(); # Add 0 to treat it as numeric (to avoid quoting)
+  $population->{allele_count} = 0 + $allele->count() if defined $allele->count(); # Add 0 to treat it as numeric (to avoid quoting)
   $population->{allele} = $allele->allele();
   $population->{submission_id} = $allele->subsnp() if $allele->subsnp();
 
--- a/lib/EnsEMBL/REST/Model/Variation.pm
+++ b/lib/EnsEMBL/REST/Model/Variation.pm
@@ -330,7 +330,9 @@ sub pops_as_hash {
   my $population;
   $population->{frequency} = 0 + $allele->frequency(); # Add 0 to treat it as numeric (to avoid quoting)
   $population->{population} = $allele->population->name();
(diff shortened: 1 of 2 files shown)


lib/EnsEMBL/REST/Model/Variation.pm
```diff
@@ -330,7 +330,7 @@ sub pops_as_hash {
   my $population;
   $population->{frequency} = 0 + $allele->frequency(); # Add 0 to treat it as numeric (to avoid quoting)
   $population->{population} = $allele->population->name();
-  $population->{allele_count} = 0 + $allele->count(); # Add 0 to treat it as numeric (to avoid quoting)
+  $population->{allele_count} = 0 + $allele->count() if defined $allele->count(); # Add 0 to treat it as numeric (to avoid quoting)
```
mkszuba: I would really recommend using a block rather than postfix form of if here. It might be slightly more verbose but at least it makes it one hundred percent clear that it is the whole assignment to $population->{allele_count} that is controlled by the condition.

helensch: Modified to use block if

t/variation.t
```diff
@@ -105,6 +105,47 @@ my $expected_result = { rs142276873 => $expected_variation_1, rs67521280 => $exp
...
+        'strand' => 1,
+        'coord_system' => 'chromosome',
+        'allele_string' => 'C/T',
+        'start' => 25744274
+     }
+    ],
+    'name' => 'COSM946275',
+    'MAF' => undef,
+    'ambiguity' => 'Y',
+    'populations' => [
```
nerdstrike: Is this susceptible to random order changes?

mkszuba: Unless I have misunderstood https://metacpan.org/pod/Test::Deep#HASH-COMPARISONS, it shouldn't be.

nerdstrike: 'populations' is a list generated from the Variation code. Test::Deep isn't allowed to do set operations on a list by default. @helensch Is the major allele always first, or is there some other sorting mechanism?

helensch: @nerdstrike yes the test it is susceptible to random order change in the 'populations' list.
As Iit is the populations that I want to compare I suggest doing  a cmp_deeply with a bag comparison
cmp_deeply($pops_json->{'populations'}, 
             bag(@{$expected_pops_nc->{'populations'}}), 
          "Returning population information - no allele_count");

nerdstrike: Sounds reasonable

PR #309, opened 2018-10-02
Only try to create log/PID dirs if the server is starting.

nicklangridge:
### Description

Only try to create log/PID dirs if the server is starting.
The server control script may be used for other purposes than starting the server, 

### Use case

e.g. the REST installer uses the server control script to generate an init script, at which time we should not create these dirs.

### Benefits

The installer works properly

### Possible Drawbacks

none

### Testing

Tested in my sandbox

### Changelog

Server log/PID dirs now only created on server start operations


nicklangridge:
If it's ok I'll make a PR for release/95 too

premanand17:
Thanks Nick. Please go ahead and make a PR for release/95.

PR #313, opened 2018-11-13
Pairwise cactus

muffato:
### Description

This change comes from the helpdesk ticket #308278. Only the Cactus (mouse-strains) multiple-alignment is currently allowed, but we have in the databases entries for pairwise alignments of these species as well. The functionality could already be achieved by querying the entire set of mouse-strains and filtering by `display_species`, but querying the pair of species directly (just like our other pairwise alignments) is more natural (and faster).

### Use case

Mouse-strains pairwise alignments are now allowed, as the alignments between the primary assembly and the haplotypes / patches (human and mouse).

### Benefits

Since the list is retrieved from the database, it will automatically find the alignments actually available. For instance, multiple-alignments are not available in EG databases, etc

### Possible Drawbacks

### Testing

Yes

Yes

Yes and unsure (there are many errors related to Variation and Regulation)

### Changelog

no


lib/EnsEMBL/REST/Model/GenomicAlignment.pm
```diff
@@ -76,18 +71,24 @@ sub get_alignment {
 
   #Get the compara DBAdaptor
   $c->forward('get_adaptors');
+
+  #Get method
+  my $method_type = $c->request->parameters->{method} || 'EPO';
+  my $method = $c->stash->{method_adaptor}->fetch_by_type($method_type);
```
nerdstrike: It doesn't seem like $method gets used for anything of substance, now you've renamed everything to method_type. Is that intentional?

muffato: `$method` is only used to get the "class" and nothing else.
I was actually wondering about that. To be as close as possible to the compara API, `method_type` should a string and `method` the object. But by doing that I indeed had to rename the variable in a few more places, which admittedly clutters the diff. Alternatively, to make the diff clearer, I could leave `$method` as it was and name the new variable `$method_object` ?

nerdstrike: Tell you what, move the invocation of $method->class to a more up-front test rather than a postfix so it's more obvious. The line is too long for github anyway. Is $method->class the same or different to $method->isa?

muffato: `class` is one of the fields _we_ define in the object, cf https://github.com/Ensembl/ensembl-compara/blob/release/94/modules/Bio/EnsEMBL/Compara/Method.pm#L138-L155 http://www.ensembl.org/info/docs/api/compara/compara_schema.html#method_link
It's a string that we use to tell what type of data the pipeline has produced.

nerdstrike: Are you content to leave this as is?

nerdstrike:
Bear in mind this on release/95 only. @mateuspatricio - someone might need to step in if this has to find its way to master as well.

PR #315, opened 2018-11-13
Missing translation end data for variants

ens-lgil:
### Description

Spotted that the `end` data returned by the `overlap/translation` for the type `transcript_variation` was wrong: the `start = end` is only true for SNP, but this not the case for short deletion (spread onto more than 1 amino acid).
I believe this is a very old bug.

### Use case

Example for `rs1326874757` (CACCGTCAT/-), it's mapped coordinates on ENSP00000465759 are `995-997` but on REST, the coordinates are `995-995`:
http://rest.ensembl.org/overlap/translation/ENSP00000465759?feature=transcript_variation;content-type=application/json

### Benefits

The REST endpoint will returns the correct coordinates


ens-lgil:
@nerdstrike, this fix needs to be in `master` as well: should I send a new PR on `master` then ?

nerdstrike:
Please do

PR #326, opened 2019-01-03
Updates to make tests pass for funcgen

mn1:
### Description
Updated tests, because the json responses have changed a bit.

### Changelog

This pull request does not change any endpoint. It is an update for two tests, because the json for motif features and binding matrices has changed.

The json representing a 
  - binding matrix now has a "max_position_sum" key instead of the "max_element" key and
  - a motif feature now has a transcription_factor_complex key.


nerdstrike:
Why does it appear that there is an extra column for one row, but not the previous? Otherwise I'm broadly in favour.

mn1:
Thanks, good catch Kieron!

I've removed this now.

magaliruffier:
I disagree that there are no changes to existing endpoints.
The test cases are meant to represent the output for a given endpoint. If you change the test case, that means that the same endpoint will now have a different json structure, which we need to communicate to our users.
Please add some lines for the change log

mn1:
Ok, probably best, if Ilias puts this in.

nerdstrike:
FYI you've got a few of these:
Use of uninitialized value $unit in string eq at /home/travis/build/Ensembl/ensembl-rest/lib/EnsEMBL/REST/Model/Regulatory.pm line 275.
Use of uninitialized value $unit in string eq at /home/travis/build/Ensembl/ensembl-rest/lib/EnsEMBL/REST/Model/Regulatory.pm line 280.
Use of uninitialized value $unit in string eq at /home/travis/build/Ensembl/ensembl-rest/lib/EnsEMBL/REST/Model/Regulatory.pm line 285.

I can't tell if they're intentional or not. It suggests you're doing something quirky.

ilavidas:
Just to confirm that I'm looking into it

ilavidas:
I have a fix for the uninitialised warnings but I do not have permissions to push it to this repo. Which of the following options do you prefer?
a) You approve this PR and I create a new one that will only address the uninitialised warnings issue.
b) You close this PR and I create a new one that will address all issues.

nerdstrike:
@ilavidas Please provide your patch at the first convenience

PR #328, opened 2019-01-03
Variant Recoder updated to accept SPDI format

dglemos:
Update Variant Recoder REST to accept SPDI format as input and output.


magaliruffier:
Please fill in the PR template.
If the output of any endpoint is changed, you will also need some documentation for the changelog, so we can inform our users of any changes that affect them

PR #329, opened 2019-01-03
Variant Recoder REST updated to accept SPDI format

dglemos:
### Description

Update Variant Recoder to accept SPDI format as input and output. 

### Use case

Allow the users to use SPDI notation. 

### Benefits

Users can input SPDI format and get the corresponding variant identifier, HGVS and SPDI notation. 

### Possible Drawbacks

None

### Testing

Added some tests and all passed. 

### Changelog

[variant_recoder/] Also retrieve a SPDI for a given variant identifier, HGVS or SPDI notation.


nerdstrike:
Thanks @dglemos for fixing the test. As above, your vep_spdi variable will only help standalone developers here. Production config is handled elsewhere.

PR #333, opened 2019-01-17
Combine e!g and e! REST

nicklangridge:
### Requirements

### Description

- Merge in e!g info and lookup endpoints, documentation, and example params
- Adapt lookup to use metadata db 
- Add registry and app preloading in Starman
- Fix some documentation example params for existing endpoints

### Use case

A single REST instance to serve all Ensembl divisions.

### Benefits

We can retire Ensembl Genomes REST.

### Possible Drawbacks

When all divisions are enabled we see 
- Increased memory footprint
- Slow server start time due to registry preloading (startup takes a couple of minutes)

### Testing

Tests for the new endpoints are in development and will follow in a separate PR (as agreed with Mag)

### Changelog

Merge Ensembl Genomes REST functionality to allow a single REST instance to serve all Ensembl divisions.


lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -94,7 +94,7 @@ sub species : Local : ActionClass('REST') :Args(0) { }
 
 sub species_GET :Local :Args(0) {
   my ($self, $c) = @_;
-  my $division = $c->request->param('division');
+  my $division = $c->request->param('division') || 'EnsemblVertebrates';
```
nerdstrike: Something is very up with the strains in the Travis test of info.t

nerdstrike: We're also leaning toward the  // operator these days, as it isn't fooled by numeric zeroes

lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -331,6 +331,131 @@ sub consequence_types_GET {
...
+### The following methods were ported from EG REST...
+
+sub ensgen_version : Chained('/') PathPart('info/eg_version') :
+  ActionClass('REST') : Args(0) { }
+
+sub ensgen_version_GET {
+  my ( $self, $c ) = @_;
+  $c->log()->info("Retrieving EG version from registry");
+  # lazy load the registry
+  $c->model('Registry')->_registry();
```
nerdstrike: It would be better for the lazy loading trigger to occur in the Registry model rather than here... I presume this is to avoid EG database scans blocking server start.

lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -331,6 +331,131 @@ sub consequence_types_GET {
...
+  return;
+}
+
+sub divisions : Chained('/') PathPart('info/divisions') :
+  ActionClass('REST') : Args(0) { }
+
+sub divisions_GET {
+  my ( $self, $c, $division ) = @_;
+  # lazy load the registry
+  $c->model('Registry')->_registry();
```
nerdstrike: This definitely code-smell at this point.

lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -331,6 +331,131 @@ sub consequence_types_GET {
...
+  $c->log()->info("Retrieving EG version from registry");
+  # lazy load the registry
+  $c->model('Registry')->_registry();
+  my $lookup = $c->model('Registry')->_lookup();
+  $self->status_ok( $c, entity => { version => $lookup->adaptor->data_release->ensembl_genomes_version } );
+  return;
+}
+
+sub genomes_name : Chained('/') PathPart('info/genomes') :
+  ActionClass('REST') : Args(1) { }
```
nerdstrike: It's a bit odd to have a mandatory argument but then ignore it and delegate to a follow on method.

lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -331,6 +331,131 @@ sub consequence_types_GET {
...
+sub genomes_taxonomy : Chained('/') PathPart('info/genomes/taxonomy') :
+  ActionClass('REST') : Args(1) { }
+
+sub genomes_taxonomy_GET {
+  my ( $self, $c, $taxon ) = @_;
+  # lazy load the registry
+  $c->model('Registry')->_registry();
+  my $lookup = $c->model('Registry')->_lookup();
+  my $expand = $c->request->param('expand');
+  $c->log()
```
nerdstrike: Indent failure. ->info and @{ } could be formatted to more clearly with respect to the run-on from the previous line.

lib/EnsEMBL/REST/Model/Registry.pm
```diff
@@ -361,7 +358,7 @@ sub _build_species_info {
     my $species_lc = ($species);
     push(@{$groups_lookup{$species_lc}}, $group);
 
-    if($group eq 'core' && $species !~ /Ancestral/) {
+    if($group eq 'core' && $species !~ /(Ancestral|DEFAULT)/) {
```
nerdstrike: Can I get a note here to know what DEFAULT is?

lib/EnsEMBL/REST/PreloadRegistry.pm
```diff
@@ -0,0 +1,89 @@
+
+=head1 LICENSE
+
+Copyright [2009-2014] EMBL-European Bioinformatics Institute
```
nerdstrike: Well this needs fixing

lib/EnsEMBL/REST/PreloadRegistry.pm
```diff
@@ -0,0 +1,89 @@
...
+#
+# We need to preload the registry otherwise it will be reloaded every time a 
+# new worker thread is spawned (this will be very slow for the unlucky users).
+# plack and starman servers support the -M param to specify a module to load
+# before workers are forked, e.g.
+#
+# starman --listen :9300 -MEnsEMBL::REST::PreloadRegistry ensembl-rest/ensembl_rest.psgi
+#
+
+use strict;
```
nerdstrike: Don't leave out the warnings.

lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -331,6 +331,131 @@ sub consequence_types_GET {
...
+sub genomes_name_GET {
+  my ( $self, $c, $name ) = @_;
+  # lazy load the registry
+  $c->model('Registry')->_registry();
+  my $lookup = $c->model('Registry')->_lookup();
+  my $expand = $c->request->param('expand');
+  $c->log()->info("Retrieving information about all genomes");
+  my $info = $lookup->adaptor()->fetch_by_any_name($name);
+  $c->go( 'ReturnError', 'custom', ["Genome $name not found"] )
+  unless defined $info;
```
nerdstrike: If you're determined to use the postfix, indent this please

nicklangridge:
@nerdstrike thanks for the review - I'll work through these points gradually.

nicklangridge:
> I'm a little wary of mixing functions in with class methods

There's plenty of precedent for this in other controllers - what's the preferred option?

nicklangridge:
> I don't like what you're doing with void context registry calls

Example please - not sure where this refers to.

lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -331,6 +331,131 @@ sub consequence_types_GET {
   return;
 }
 
+### The following methods were ported from EG REST...
+
+sub ensgen_version : Chained('/') PathPart('info/eg_version') :
+  ActionClass('REST') : Args(0) { }
+
+sub ensgen_version_GET {
```
nerdstrike: Apparently this isn't tested!

lib/EnsEMBL/REST/Controller/Info.pm
```diff
@@ -331,6 +331,131 @@ sub consequence_types_GET {
...
+  unless defined $info;
+  $self->status_ok( $c, entity => _expand_genome($info, $expand) );
+  return;
+}
+
+sub divisions : Chained('/') PathPart('info/divisions') :
+  ActionClass('REST') : Args(0) { }
+
+sub divisions_GET {
+  my ( $self, $c, $division ) = @_;
```
nerdstrike: Redundant

nerdstrike:
For posterity, Nick and I have found a way that might allow the EG lookup system be configured directly via Registry instead of instantiating it off the back of the REST Registry model. That would remove most of strange invocations of _lookup and _registry in these endpoints.

Renaming of the "lookup" to something more meaningful would also help to delineate between EG metadata/species information, and the stable ID lookup database.

lib/EnsEMBL/REST/Controller/Info.pm
nicklangridge: Changed to //

lib/EnsEMBL/REST/Controller/Info.pm
nicklangridge: This has gone now

lib/EnsEMBL/REST/Controller/Info.pm
nicklangridge: This has gone now

lib/EnsEMBL/REST/Controller/Info.pm
nicklangridge: Not planning to change this, unless you feel strongly and can suggest better option?

lib/EnsEMBL/REST/Controller/Info.pm
nicklangridge: Reformatted

lib/EnsEMBL/REST/Model/Registry.pm
nicklangridge: Removed

lib/EnsEMBL/REST/PreloadRegistry.pm
nicklangridge: Copyright updated

lib/EnsEMBL/REST/PreloadRegistry.pm
nicklangridge: Added warnings

lib/EnsEMBL/REST/Controller/Info.pm
nicklangridge: Changed formatting

lib/EnsEMBL/REST/Controller/Info.pm
nicklangridge: Tests will follow...

nicklangridge:
> Renaming of the "lookup" to something more meaningful would also help to delineate between EG metadata/species information, and the stable ID lookup database.

The lookup property on the Registry model is no longer needed and has been removed - instead the genome info adaptor is provided directly by the registry.

lib/EnsEMBL/REST/Controller/Info.pm
nicklangridge: Removed

lib/EnsEMBL/REST/Controller/Info.pm
nerdstrike: Fair enough... I'm not prepared to declare a one true way right now. Catalyst is too finicky around these decorators.

lib/EnsEMBL/REST/PreloadRegistry.pm
```diff
@@ -11,7 +11,7 @@ You may obtain a copy of the License at
 
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+WITHinfo WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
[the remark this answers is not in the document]
nicklangridge: fixed

lib/EnsEMBL/REST/PreloadRegistry.pm
```diff
@@ -0,0 +1,100 @@
...
+
+my @db_servers;
+
+# look for primary db server
+if ($reg_config->{host} && $reg_config->{user} && $reg_config->{port}) {
+  info("Using db host '$reg_config->{host}'");
+  push @db_servers, {
+    -host => $reg_config->{host}, 
+    -user => $reg_config->{user}, 
+    -port => $reg_config->{port}
```
nerdstrike: Do you think this represents sufficient options for every use?

nicklangridge: Sufficient for existing usecases - do you see potential problems?

nicklangridge: Could easily add password - anything else needed?

nerdstrike: I'm used to having to expose all the registry vars, but most are redundant in production.  It's fine to leave as is if this module only fires when you're being clever with starman options.

lib/EnsEMBL/REST/PreloadRegistry.pm
```diff
@@ -0,0 +1,100 @@
...
+info('Building species info...');
+
+EnsEMBL::REST::RegistryHelper::build_species_info();
+
+info('Done');
+
+
+sub info {
+  my $msg = shift;
+  warn sprintf("[%s] %s\n", __PACKAGE__, $msg);
```
nerdstrike: warn already mentions what file the warn comes from. \n is redundant too.

lib/EnsEMBL/REST/PreloadRegistry.pm
nerdstrike: I see you've baked it into server_control.pl . Not being able to specify passwords and db_version might impact developer testing.

lib/EnsEMBL/REST/PreloadRegistry.pm
nicklangridge: \n tells warn not to report the filename/line number etc - so here only the package prefix and the message is shown - it's less noisy

lib/EnsEMBL/REST/PreloadRegistry.pm
nicklangridge: server control is in a folder called `bin/production`, is it intended for developer use?

nerdstrike: Mostly not, but sometimes somebody adds HDF5 to the mix and the multi-thread behaviour matters

nerdstrike:
b.t.w. I'm still seeing 
# *  16|      strain => 'test_strain',           *    |                                         |
# *  17|      strain_collection => 'human',      *    |                                         |

in the info.t failures. I've forgotten what our resolution was.

lib/EnsEMBL/REST/PreloadRegistry.pm
nicklangridge: ok, i'll add the extra params

PR #335, opened 2019-01-28
ENSCORESW-3015: replace multi with vertebrates for compara usage

magaliruffier:
### Requirements

### Description

_Using one or more sentences, describe in detail the proposed changes._
The default compara is set to 'vertebrates' instead of 'multi. Under the hood, 'multi' is still allowed for backwards compatibility.

### Use case

_Describe the problem. Please provide an example representing the motivation behind the need for having these changes in place._
For 96, we will have a single REST server providing access to all Ensembl divisions. For consistency across divisions, we do not want to have a 'multi' compara database that refers to vertebrates, when all other comparas are identified by their divisions (eg Plants, Fungi...)
This change allows us to present all divisions equally to our users while ensuring existing functionality remains unchanged.
In particular, the default compara is set to 'vertebrates'. That way, users who did not have to specify a compara database until now because only one was available (which is the case for anyone using vertebrates until release 95) will not have to change any of their queries.

### Benefits

_If applicable, describe the advantages the changes will have._
When querying the /info/comparas endpoint, users will get an meaningful list of compara databases available based on their division name.
This will also match the output of the /info/divisions endpoint.
Any other functionality will remain unchanged.

### Possible Drawbacks

'multi' can still be used as an equivalent of 'vertebrates', exposing some of our internal workings

### Testing

Yes

Yes

Yes

### Changelog

No change of functionality

### DIFF
--- a/lib/EnsEMBL/REST/Controller/GenomicAlignment.pm
+++ b/lib/EnsEMBL/REST/Controller/GenomicAlignment.pm
@@ -25,7 +25,7 @@ require EnsEMBL::REST;
 
 BEGIN { extends 'Catalyst::Controller::REST'; }
 
-has 'default_compara' => ( is => 'ro', isa => 'Str', default => 'multi' );
+has 'default_compara' => ( is => 'ro', isa => 'Str', default => 'vertebrates' );
 has 'compara_base_dir_location' => ( is => 'ro', isa => 'Str' );
 has 'max_slice_length' => ( isa => 'Num', is => 'ro', default => 1e6);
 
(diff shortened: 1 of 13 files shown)


magaliruffier:
Travis failure is unrelated to proposed change, as it is still failing on the ga4gh tests

magaliruffier:
Your interpretation is correct.
Fair point about the lower case, I think it would be taken care of by the get_best_compara method, but it does not harm to be extra careful, I've updated the comparison
I am not aware of any more changes to this for 97, we are going for minimal change to both user experience and internal API workings

nicklangridge:
This looks ok to me - by eye I can't spot any cases you haven't covered.

magaliruffier:
I have updated the if statements in multi-liners
I am not polluting this PR with cosmetic changes to otherwise untouched code

PR #337, opened 2019-02-08
Feature/fix strain info

nerdstrike:
### Description

The addition of a default division to the species/info endpoint (EnsemblVertebrates) has exposed a bug in the Model::Registry->get_species_info, which was made without the expectation of divisions for everything

### Use case

Strains hidden for vertebrates so that users are not surprised with new seemingly unimportant keys in the returned documents.

### Benefits

Code is cleaner, works as intended, and is able to filter on division AND strain collection, not one or the other.

### Possible Drawbacks

None obvious

### Testing

Testing environment required more specific hacks for BioPerl versions to protect Bio::DB::HTS. Tests are otherwise unchanged outside of some whitespace.

Pass

Indeed

### Changelog

Not relevant to users.

### DIFF
--- a/lib/EnsEMBL/REST/Model/Registry.pm
+++ b/lib/EnsEMBL/REST/Model/Registry.pm
@@ -219,27 +219,36 @@ sub get_species {
   my ($self, $division, $strain_collection, $hide_strain_info) = @_;
   my $info = $self->_species_info();
   #have to force numerification again. It got lost ... somewhere
-  #filter by division (e.g.: EnsemblVertebrates)
-  return [ grep { $_ > 0 } map {$_->{release} += 0; $_} grep { lc($_->{division}) eq lc($division) } @{$info}] if $division;
-
-  #filter by strain_collection (e.g.: mouse)
-  return [ grep { $_ > 0 } map {$_->{release} += 0; $_} grep { lc($_->{strain_collection}) eq lc($strain_collection) } @{$info}] if $strain_collection;
-
   #flag to show or hide strain info
-  if($hide_strain_info > 0){
+  if( $hide_strain_info > 0 ){
     my @hidden_keys = qw(strain strain_collection);
     my $new_info_list = [];
-    foreach my $inf(@{$info}){
-    my $new_info_hash = {};
-      foreach my $key(keys %$inf){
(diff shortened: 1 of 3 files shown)


t/info.t
```diff
@@ -75,33 +75,79 @@ is_json_GET(
...
+    $get_species->('/info/species', 'Checking only DBA available is the test DBA'),
+    $expected, 
+    "/info/species | Checking only DBA available is the test DBA"
+  );
   eq_or_diff_data(
-    $get_species->('/info/species?division=EnsemblVertebrates', q{Output is same as /info/species if specified 'EnsemblVertebrates' division}), $expected, 
-    "/info/species?division=EnsemblVertebrates | Output is same as /info/species if specified 'Ensembl' division");
+    $get_species->('/info/species?division=EnsemblVertebrates', q{Output is same as /info/species if specified 'EnsemblVertebrates' division}),
+    $expected, 
+    "/info/species?division=EnsemblVertebrates | Output is same as /info/species if specified 'Ensembl' division"
```
premanand17: Shouldn't the message be "......if specified 'EnsemblVertebrates' division" ?!

nerdstrike: Agreed

lib/EnsEMBL/REST/Model/Registry.pm
```diff
@@ -219,27 +219,36 @@ sub get_species {
...
-    push (@$new_info_list, $new_info_hash);
+      push @$new_info_list, $new_info_hash;
     }
-    return $new_info_list;
+    $info = $new_info_list;
   }
-
-  return [ grep { $_ > 0 } map {$_->{release} += 0; $_} @{$info} ];
+  
+  my $filter;
```
premanand17: Is $filter used anywhere below?

nerdstrike: Nope, it was an aborted attempt at cleaning it.

nerdstrike:
Nick has already changed the default options on master, so it's a little clearer. We can have a proper review of docs if we need to improve them for clarity. It's been quite a few years since we did one of those.

muffato:
Thanks ! That will make our Travis happy

PR #338, opened 2019-02-08
Python3 print needs parentheses

MatBarba:
### Description

Add parentheses to the "print" call in the rest API documentation for Python 3 (in case of plain text result).

### Use case

Python3 needs parentheses for "print", contrary to Python2. Users copying-pasting the example code are currently getting an syntax error.

### Benefits

No syntax error for users...

### Possible Drawbacks

None.

### Testing

No tests changed: only a doc template was modified.


MatBarba:
I'll recreate this PR against master to make it cleaner...

PR #341, opened 2019-02-19
Fixes for joint REST

nicklangridge:
### Description

A few fixes for joint REST
- Fix `info/info_genome` endpoint
- Use better format for specifying multiple db configs 
- Fix registry helper bug by removing a db disconnect left in by mistake

### DIFF
--- a/lib/EnsEMBL/REST/RegistryHelper.pm
+++ b/lib/EnsEMBL/REST/RegistryHelper.pm
@@ -137,7 +137,6 @@ sub build_species_info {
         $processed_db{$db_key} = 1;
       }
     }
-    $dba->dbc->db_handle->disconnect;
   }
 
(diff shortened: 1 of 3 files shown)


lib/EnsEMBL/REST/RegistryHelper.pm
```diff
@@ -137,7 +137,6 @@ sub build_species_info {
         $processed_db{$db_key} = 1;
       }
     }
-    $dba->dbc->db_handle->disconnect;
```
nerdstrike: I'm a bit wary of this, but I realise you're managing the connections in another way.

nicklangridge: I added that line originally to fix an situation that no longer arises - it should not have made it into the previous PR

PR #346, opened 2019-03-21
Update Beacon version

dglemos:
### Description

Update version number of Beacon endpoint 

### Use case

Endpoint had an older version. 

### Benefits

Updated endpoint 

### Possible Drawbacks

None

### Testing

No 

No 

### Changelog

No


nerdstrike:
Do I understand that the version of the beacon was wrong? Or that version 0.3 is the same as version 1 as far as we are concerned?

dglemos:
It's just outdated, but I need to make more changes to this pull request that I didn't realize before. I'm going to close it.

PR #358, opened 2019-06-12
Add missing YAML CPAN dependency

nicklangridge:
Add missing YAML CPAN dependency as required here https://github.com/Ensembl/ensembl-rest/blame/release/97/lib/EnsEMBL/REST/Model/Documentation.pm#L33

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -19,6 +19,7 @@ requires 'Catalyst::View::JSON';
 requires 'Catalyst::View::TT';
 requires 'JSON::XS';
 requires 'JSON';
+requires 'YAML';
 requires 'YAML::XS';
 requires 'YAML::Syck';
 requires 'URI::Find';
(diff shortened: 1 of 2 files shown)


nerdstrike:
Odd. Does YAML::XS not imply YAML vanilla?

mkszuba:
As far as I can tell from metacpan, not yet - the author plans to make YAML a front-end module for various implementations but for now it's just a specific, pure-Perl implementation - and a deprecated one to boot.

nicklangridge:
Apparently not https://metacpan.org/pod/YAML::XS
> Dependencies
> none

I had to install it manually for the REST server this release. I'm not sure I've had to do it previously, but given that the YAML module is explicitly used I think it should be in cpanfile.

nicklangridge:
I guess the use of YAML could be switched to YAML::XS but I'll leave that for you to decide.

mkszuba:
@nicklangridge Just a quick reminder that there isn't much time left before the branching of release 98, if you would like to have this fixed by then.

nicklangridge:
So you want a PR with YAML::XS in Documentation.pm instead of this solution?

mkszuba:
Yes, if it works of course.

PR #359, opened 2019-06-17
Test DB: add missing ontology patches

mkszuba:
The last patching of test databases in this repository missed ontology-DB patches because the patcher script has only just been updated to look for said patches in their new home. Make sure test databases are up to date before the release date.


mkszuba:
Test failures are due to an unrelated issue, merging.

PR #369, opened 2019-07-18
correction for allowed region size

at7:
Same as https://github.com/Ensembl/ensembl-rest/pull/368

### Description
Documentation update. Correct the maximum allowed region size. We added a check to the code for restricting the maximum allowed region size a long time ago but forgot to update the documentation. This has been reported by a user on the ensembl-dev list.

### Use case

NA

### Benefits

Correct documentation

### Possible Drawbacks

None

### Testing

Not needed

### Changelog

Not needed

### DIFF
--- a/root/documentation/ld.conf
+++ b/root/documentation/ld.conf
@@ -98,7 +98,7 @@
       </species>
       <region>
         type=String
-        description=Query region. A maximum of 1Mb is allowed.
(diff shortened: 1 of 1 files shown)


ens-bwalts:
Test failures  due to incompatible test dbs. Change is documentation-only, does not affect functionality, and change passes on release/98. Will go ahead and merge.

PR #372, opened 2019-07-26
Bugfix/ld in mhc region

at7:
### Description

Helpdesk reported a problem with getting LD in the MHC region. Trying to use the LD region endpoint in the MHC region causes timeouts caused by the high density of variation in that region.  We are now checking the input region and throw an exception if it overlaps the MHC region and only allow a maximum region size of 10KB instead of 500KB. I tested it with my local REST installation. I tried 100KB and 50KB but even 50 was still taking a long time to return results.

### Use case

We updated the LD region endpoint which calculates pairwise LD for all variants in a given region. The MHC region has a very high density of variation and pairwise calculation can get very big and cause runtime and memory problems.

### Benefits

Fewer problems for the REST server. Requests with default region size of 500KB in the MHC region causes memory problems and timeouts.

### Possible Drawbacks
None .
Users will most likely have problems computing LD in MHC region with current settings. Getting back an error message if the selected region overlaps MHC and is too big is a good compromise. At least users know why the endpoint has problems.

### Testing

I added tests to make sure that the overlap with MCH region is reported correctly.

Only archive,t fails when running entire test suite.

### Changelog

Would you consider this as a change in functionality? I'm happy to add a comment to the changelog.


mkszuba:
@at7 Were likely timeouts an expected, documented feature of this code? If not, not a change of functionality IMHO.

lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
```diff
@@ -81,6 +81,10 @@ sub fetch_LDFeatureContainer_variation_name {
 sub fetch_LDFeatureContainer_slice {
   my ($self, $slice, $population_name) = @_;
   Catalyst::Exception->throw("No region given. Please specify a region to retrieve from this service.") if ! $slice;
+  if (slice_overlaps_mhc_region($slice)) {
+    my $mhc_region = get_mhc_region($slice);
+    Catalyst::Exception->throw("Specified region overlaps MHC region $mhc_region and can therefore not be greater than 10KB.") if ($slice->length > 10_000);
```
mkszuba: Please change this if statement to block syntax. I know that postfix syntax is generally tolerated for terminal statements like this one but this line is so long that having to look all the way to the right to see that it is conditional really breaks the flow of reading the code.
Consider changing this in lines: 83 and88 as well while at it, seeing as they are so close to a new code block.

lib/EnsEMBL/REST/Model/LDFeatureContainer.pm
```diff
@@ -219,6 +223,38 @@ sub to_array {
...
+    return 1 if ($start > $mhc_start && $start < $mhc_end);
+    return 1 if ($end > $mhc_start && $end < $mhc_end);
+  } 
+  return 0; 
+}
+
+sub get_mhc_region {
+  my $slice = shift;
+  my $assembly = $slice->coord_system->version;
+  return '' unless (defined $mhc_regions->{$assembly});
```
mkszuba: I would recommend using q{} instead of '' here.

t/ld.t
```diff
@@ -199,6 +199,15 @@ cmp_bag($json, $expected_output, "Example region, population, r2");
...
 
+$ld_region_get = '/ld/homo_sapiens/region/6:28510120..28610120/1000GENOMES:phase_1_ASW?r2=0.5';
+action_bad($ld_region_get, 'Specified region overlaps MHC region');
+# partially overlaps beginning of MHC region
+$ld_region_get = '/ld/homo_sapiens/region/6:28410120..28610120/1000GENOMES:phase_1_ASW?r2=0.5';
+action_bad($ld_region_get, 'Specified region overlaps MHC region');
+# partially overlaps end of MHC region
+$ld_region_get = '/ld/homo_sapiens/region/6:33438354..33458354/1000GENOMES:phase_1_ASW?r2=0.5';
+action_bad($ld_region_get, 'Specified region overlaps MHC region');
+
```
ens-bwalts: Could you please add a positive test for fetching in the MHC region (test to see if fetching a region < 10kb succeeds).

at7:
Thank you to both of you for the feedback

PR #373, opened 2019-07-26
Bump core and ontology test databases to schema 99

mkszuba:
### Description

Bump the schema version of test core and ontology databases.

### Use case

Consequence of the fact we now bump the Core API version, and by consequence Core and ontology database-schema versions, very early in the release cycle. Several tests most notably in archive.t, fail on master if ensembl-rest test databases are not patched accordingly.

### Benefits

archive.t tests will stop failing.

### Possible Drawbacks

This will have to be done for every release if we intend to stand by bumping core version early in the release cycle.

### Testing

No.

N/A

Yes, archive.t tests have stopped failing and haven't seen any new failures crop up.


mkszuba:
Argh. I have misread the output of runtests on my machine, there ARE new errors... and this time they are related to the Variation schema. Early version bump is becoming more and and more unpractical unless it is rolled out across the whole of Ensembl.

PR #375, opened 2019-08-08
Model::Lookup::find_object_location(): only pass the expected three substitution arguments to sprintf()

mkszuba:
### Description

Make sure that the sprintf() call in line 350 of _lib/EnsEMBL/REST/Model/Lookup.pm_ doesn't receive more than the 3 substitution arguments it expects.

### Use case

Previously said sprintf() call was passed the entire @captures array and at least nowadays that array is longer than 3 elements, resulting in a "redundant arguments" warning which spams the output of the test suite.

### Benefits

Less noise in test output.

### Possible Drawbacks

None I can think of.

### Testing

Yes, _archive.t_ - the first test file to have previously produced this warning - now uses _Test::Warnings_. Note that this adds a new test-time dependency to _ensembl-rest_, which however shouldn't be a problem because other Ensembl repositories already depend on _Test::Warnings_.

The test passes _i.e._ no warnings are detected.

Yes, no regression detected. The errors in _archive.t_ are the same as before with the bugfix branch based on _master_ and gone when based on the branch of #374.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -69,3 +69,4 @@ test_requires 'HTTP::Request::Common';
 test_requires 'Plack::Test';
 test_requires 'Net::CIDR::Lite';
 test_requires 'Test::Time::HiRes';
+test_requires 'Test::Warnings';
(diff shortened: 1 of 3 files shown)


mkszuba:
Huh, I could have sworn I described benefits and drawbacks when I created this pull request. Fixed now.

nerdstrike:
Thanks for finding this bug, it has annoyed me on and off for years.

PR #396, opened 2019-10-15
Add option to include SO consequence ranking

dglemos:
### Description

Add an option to include the consequence ranking for the sequence ontology terms. 

### Use case

There was a request to add the consequence rank for the SO terms. 

### Benefits

One more option to get more info from the consequence terms. 

### Possible Drawbacks

None. This change doesn't affect old queries. 

### Testing

Yes

The tests pass. 

No regression detected. 

### Changelog

No, the functionality is the same there's only one more option. 
/info/variation/consequence_types

### DIFF
--- a/lib/EnsEMBL/REST/Model/Variation.pm
+++ b/lib/EnsEMBL/REST/Model/Variation.pm
@@ -465,20 +465,33 @@ sub fetch_population_infos {
 }
 
 sub fetch_consequence_types {
-  my ($self) = @_;
+  my ($self, $rank) = @_;
 
   my @consequence_types = ();
 
   my $oc;
   foreach my $key(keys %OVERLAP_CONSEQUENCES) {
     $oc = $OVERLAP_CONSEQUENCES{$key};
-    push @consequence_types,
+
+    if (defined $rank && $rank == 1) {
+        push @consequence_types,
+          {
+            SO_term => $oc->SO_term,
+            SO_accession => $oc->SO_accession,
+            label => $oc->label,
+            description =>  $oc->description,
+            consequence_ranking => $oc->rank
+          }
+    }
+    else {
+      push @consequence_types,
         {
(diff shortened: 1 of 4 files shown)


lib/EnsEMBL/REST/Model/Variation.pm
```diff
@@ -465,20 +465,33 @@ sub fetch_population_infos {
...
+    }
+    else {
+      push @consequence_types,
         {
           SO_term => $oc->SO_term,
           SO_accession => $oc->SO_accession,
           label => $oc->label,
           description =>  $oc->description
         }
+    }
```
tgrego: The duplication should be avoided by setting the default stuff, and then adding the specifics... something like:
```
my $item = {
  SO_term      => $oc->SO_term,
  SO_accession => $oc->SO_accession,
  label        => $oc->label,
  description  => $oc->description
};

if ($rank) {
  $item->{'consequence_ranking'} = $oc->rank;
}

push @consequence_types, $item;
```

lib/EnsEMBL/REST/Model/Variation.pm
```diff
@@ -465,20 +465,33 @@ sub fetch_population_infos {
...
+  my ($self, $rank) = @_;
 
   my @consequence_types = ();
 
   my $oc;
   foreach my $key(keys %OVERLAP_CONSEQUENCES) {
     $oc = $OVERLAP_CONSEQUENCES{$key};
-    push @consequence_types,
+
+    if (defined $rank && $rank == 1) {
```
tgrego: `$rank` is a boolean, so we can just check `if ($rank)`

lib/EnsEMBL/REST/Model/Variation.pm
dglemos: Thanks for the suggestions.

PR #399, opened 2019-11-12
Compara endpoint update for grch37

carlacummins:
### Description

Updated compara endpoints to support removal of many species and analyses from GRCh37 REST service. This includes moving default values to the conf file (instead of hardcoding in the endpoint) and updating documentation & examples for GRCh37 purposes

### DIFF
--- a/ensembl_rest.conf.default
+++ b/ensembl_rest.conf.default
@@ -151,9 +151,16 @@ jsonp=1
     ontology_term_id=GO:0005667
     ontology_term_name=transcription factor complex
     
-    compara=multi
+    compara=vertebrates
     compara_method=EPO
     compara_method_type=GenomicAlign
+    compara_hom_format=full
+    compara_hom_type=all
+    compara_aligned=1
+    compara_sequence=protein
+    compara_cigar_line=1
+    compara_species_set_group=mammals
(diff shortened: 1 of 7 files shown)


muffato:
Travis is now failing because the alignment endpoint is now returning unaligned sequences, and other output changes in the homology endpoints. If the default values have changed they need to be listed in order to be included in the Changelog and the test-suite needs to be updated. Otherwise, change the default values to what they were to preserve compatibility

carlacummins:
@ameya1981 : which ensembl_rest.conf file is used by Travis? I didn't change the default values, I just moved them out to the config file.

magaliruffier:
> @ameya1981 : which ensembl_rest.conf file is used by Travis? I didn't intend to change the default values, I just moved them out to the config file.

ensembl_rest_testing.conf, for example
https://github.com/Ensembl/ensembl-rest/blob/release/98/ensembl_rest_testing.conf

carlacummins:
Failing tests are now in non-compara endpoints. Advice on how to proceed would be appreciated - should I try to fix them in this branch? (if yes, how?) Or are there fixes elsewhere that should be pulled into this branch? Thanks

PR #400, opened 2019-11-22
Update Beacon allele response

dglemos:
### Description

Update Beacon endpoint: small change to the response, fix a bug and add a test.

### Use case

This update is necessary to have a more precise response. 

### Benefits

Improve beacon response.

### Possible Drawbacks

None, the old queries still work and the output has the same format. 

### Testing

Yes. Update one test and add a new one. 

Pass

No regression detected. 

### Changelog

The functionality is the same.

### DIFF
--- a/lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
+++ b/lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
@@ -553,8 +553,6 @@ sub variant_exists {
         next;
     }
 
-    print "$start_pos-$end_pos, $new_start-$new_end\n";
-
     # Precise match for snv or small indels - end does not make a difference
     if ($sv == 0 && ($seq_region_start != $start_pos) && ($seq_region_end != $end_pos)) {
       next;
@@ -757,9 +755,9 @@ sub get_dataset_allele_response {
         }
 
         # Checks if dataset_id is one of the datasets where variant is found
-        # If it's not found then dataset response won't include variant
-        my %x = %{$variant_dt};
-        my $datasets = $x{$var_name};
+        # If it's not found then dataset response won't include this variant
+        my %var_dataset = %{$variant_dt};
+        my $datasets = $var_dataset{$var_name};
         my $contains = contains_value($dataset_id, $datasets);
         if($contains) {
           $count += 1;
--- a/lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
+++ b/lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
@@ -559,7 +559,8 @@ sub variant_exists {
     }
     # Match for structural variants
     if($sv == 1){
-      next unless ($seq_region_start >= $start_pos && $seq_region_start <= $start_max_pos && $seq_region_end >= $end_min_pos && $seq_region_end <= $end_pos);
+      next if ($seq_region_start < $start_pos || $seq_region_start > $start_max_pos
+              || $seq_region_end < $end_min_pos || $seq_region_end > $end_pos);
     }
 
     # Variant is a SNV
@@ -742,10 +743,9 @@ sub get_dataset_allele_response {
       $externalURL = "http://grch37.ensembl.org";
     }
 
-    my $count = 0;
+    my @urls;
 
     if($vf) {
-      my $url = '';
       my $var_name;
       my $delimiter;
       foreach my $variant (@{$vf}) {
@@ -764,15 +764,15 @@ sub get_dataset_allele_response {
         my $datasets = $var_dataset{$var_name};
         my $contains = contains_value($dataset_id, $datasets);
         if($contains) {
-          $count += 1;
-          $url .= ', ' . $externalURL . "/Homo_sapiens/" . $delimiter . $var_name;
+          my $url_tmp = $externalURL . "/Homo_sapiens/" . $delimiter . $var_name;
+          push @urls, $url_tmp;
         }
       }
-      $url =~ s/, //;
+      my $url = join(',', @urls);
       $externalURL = $url;
     }
 
-    $ds_response->{'variantCount'} = $count;
+    $ds_response->{'variantCount'} = scalar @urls;
 
     $ds_response->{'externalUrl'} = $externalURL;
 
--- a/lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
+++ b/lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
@@ -93,15 +93,11 @@ sub get_beacon_organization {
 
   my $organization;
  
-  my $description = "The European Bioinformatics Institute (EMBL-EBI)
-                      is part of EMBL, an international, innovative
-                      and interdisciplinary research organisation funded
-                      by 26 member states and two associate member states
-                      to provide the infrastructure needed to share data
-                      openly in the life sciences.";
+  my $description = "The European Bioinformatics Institute (EMBL-EBI) is part of EMBL, an international, innovative "
+                      . "and interdisciplinary research organisation funded by 26 member states and two associate member states "
+                      . "to provide the infrastructure needed to share data openly in the life sciences.";
 
-  my $address = "EMBL-EBI, Wellcome Genome Campus, Hinxton,
-                  Cambridgeshire, CB10 1SD, UK";
+  my $address = "EMBL-EBI, Wellcome Genome Campus, Hinxton, Cambridgeshire, CB10 1SD, UK";
 
   # The welcome URL depends on the assembly requested
   #my $db_assembly = $db_meta->{assembly};
@@ -782,14 +778,13 @@ sub get_dataset_allele_response {
 sub contains_value {
   my ($value, $array) = @_;
 
-  my $related;
   foreach my $i ($array) {
     if(grep { $_ eq $value } @{$i}) {
       return 1;
     }
   }
 
-  return $related;
+  return;
 }
 
 with 'EnsEMBL::REST::Role::Content';
--- a/lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
+++ b/lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
@@ -92,11 +92,13 @@ sub get_beacon_organization {
   my ($self) = @_;
 
   my $organization;
-  
-  my $description = "Ensembl creates, integrates and distributes reference datasets"
-                    . " and analysis tools that enable genomics."
-                    . " We are based at EMBL-EBI and our software"
-                    . " and data are freely available.";
+ 
+  my $description = "The European Bioinformatics Institute (EMBL-EBI)"
+                    . " is part of EMBL, an international, innovative" 
+                    . " and interdisciplinary research organisation funded"
+                    . " by 26 member states and two associate member states"
+                    . " to provide the infrastructure needed to share data"
+                    . " openly in the life sciences.";
 
   my $address = "EMBL-EBI, Wellcome Genome Campus, Hinxton, "
                 . "Cambridgeshire, CB10 1SD, UK";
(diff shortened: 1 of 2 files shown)


lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
```diff
@@ -541,61 +553,77 @@ sub variant_exists {
...
     if ($sv == 0) {
     $ref_allele_string = $vf->ref_allele_string();
     $alt_alleles = $vf->alt_alleles();
 
-    if ((uc($ref_allele_string) eq uc($new_ref))
-      && (grep(/^$new_alt$/i, @{$alt_alleles}))) {
+    # Some ins/del have allele string 'G/-', others have TG/T
+    # trim sequence doesn't always work
+    if ((uc($ref_allele_string) eq uc($ref_allele) || (uc($ref_allele_string) eq uc($new_ref)))
+      && (grep(/^$alt_allele$/i, @{$alt_alleles}) || grep(/^$new_alt$/i, @{$alt_alleles}))) {
```
tgrego: please use the block form of grep, i.e `grep { /^$alt_allele$/i } @{$alt_alleles}`
perhaps also update the string boundaries to \A \z instead of ^ $.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
```diff
@@ -705,17 +738,54 @@ sub get_dataset_allele_response {
...
 
   return $ds_response;
 }
 
+sub contains_value {
+  my ($value, $array) = @_;
+
+  my $related; 
+  foreach my $i ($array){ 
+    if (grep $_ eq $value, @{$i}) { $related = 1; }
```
tgrego: please use the block form of grep

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
```diff
@@ -705,17 +738,54 @@ sub get_dataset_allele_response {
...
+        # If it's not found then dataset response won't include this variant
+        my %var_dataset = %{$variant_dt};
+        my $datasets = $var_dataset{$var_name};
+        my $contains = contains_value($dataset_id, $datasets);
+        if($contains) {
+          $count += 1;
+          $url .= ', ' . $externalURL . "/Homo_sapiens/" . $delimiter . $var_name;
+        }
+      }
+      $url =~ s/, //;
```
tgrego: why are we adding ', ' to the $url just to remove it later?
am I missing something?

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
```diff
@@ -541,61 +553,77 @@ sub variant_exists {
...
+
         if ($incl_ds_response) {
-          $vf_found = $vf;
+          push (@vf_found, $vf);
         }
-        last;
-      } else {
-          $found = 0;
       }
     }
```
tgrego: something is wrong with the indentation here... please fix

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
```diff
@@ -541,61 +553,77 @@ sub variant_exists {
...
     # Match for structural variants
     if($sv == 1){
-      next unless ($seq_region_start >= $new_start && $seq_region_start <= $start_max_pos && $seq_region_end >= $end_min_pos && $seq_region_end <= $new_end);
-    }
-
-    # All datasets where variant was found
-    my $datasets_found = $vf->get_all_VariationSets();
-    foreach my $set (@$datasets_found) {
-      $dataset_var_found{$set->dbID()} = $set;
+      next unless ($seq_region_start >= $start_pos && $seq_region_start <= $start_max_pos && $seq_region_end >= $end_min_pos && $seq_region_end <= $end_pos);
```
tgrego: unless is to be avoided.
unless >= is equivalent to if <, just a lot harder to read.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
```diff
@@ -541,61 +553,77 @@ sub variant_exists {
...
+        'INV'    => 'inversion',
         'DUP:TANDEM' => 'tandem_duplication'
       );
 
       my $vf_so_term;
       $so_term = defined $terms{$alt_allele} ? $terms{$alt_allele} : $alt_allele;
       $vf_so_term = $vf->class_SO_term();
 
-      if ($vf_so_term eq $so_term) {
+      if ($so_term =~ /$vf_so_term/) {
```
tgrego: do we really want a regex here? CNV and contains deletion and duplication...
and please update the regex to use ${} for variables. it's now very confusing because you also use $ as start of string.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
```diff
@@ -368,20 +370,24 @@ sub get_beacon_allele_request {
...
-    $beaconAlleleRequest->{startMax} = $data->{startMax};
+    $beaconAlleleRequest->{startMax} = $data->{startMax} + 0;
   }
   if (exists $data->{endMin}) {
-    $beaconAlleleRequest->{endMin} = $data->{endMin};
+    $beaconAlleleRequest->{endMin} = $data->{endMin} + 0;
   }
   if (exists $data->{endMax}) {
-    $beaconAlleleRequest->{endMax} = $data->{endMax};
+    $beaconAlleleRequest->{endMax} = $data->{endMax} + 0;
```
tgrego: why do you +0 here?

perhaps $data->{whatever} can be undef?
in that case would it be better to  // 0 ?

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
nerdstrike: You could make an early exit here, and `return 1` instead of storing 1 until the grep is done with all the lists.

t/gabeacon.t
```diff
@@ -497,6 +497,25 @@ my $expected_data8 = {
   "datasetAlleleResponses" => undef
 };
 
+my $allele_request_2 = {
```
nerdstrike: I see you do not test your URL builder above.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
```diff
@@ -92,11 +92,13 @@ sub get_beacon_organization {
   my ($self) = @_;
 
   my $organization;
-  
-  my $description = "Ensembl creates, integrates and distributes reference datasets"
-                    . " and analysis tools that enable genomics."
-                    . " We are based at EMBL-EBI and our software"
-                    . " and data are freely available.";
+ 
+  my $description = "The European Bioinformatics Institute (EMBL-EBI)"
```
nerdstrike: FWIW heredocs exist for multiline text blocks. Less quoting needed!

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
tgrego: oh I see now...
@nerdstrike is correct, building a `@url` would be a better way of doing this, and would remove the need for `$count` (that's the size of the array now) and just join it in the end to build the mega url string.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
tgrego: what you've just done is not what you want to do...

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
```diff
@@ -705,17 +743,55 @@ sub get_dataset_allele_response {
...
+
     $ds_response->{'externalUrl'} = $externalURL;
 
   return $ds_response;
 }
 
+sub contains_value {
+  my ($value, $array) = @_;
+
+  my $related;
```
nerdstrike: This is now redundant and you can `return;`

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
```diff
@@ -92,14 +92,16 @@ sub get_beacon_organization {
   my ($self) = @_;
 
   my $organization;
-  
-  my $description = "Ensembl creates, integrates and distributes reference datasets"
-                    . " and analysis tools that enable genomics."
-                    . " We are based at EMBL-EBI and our software"
-                    . " and data are freely available.";
+ 
+  my $description = "The European Bioinformatics Institute (EMBL-EBI)
```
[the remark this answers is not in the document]
dglemos: I would prefer to leave it as it was before.

t/gabeacon.t
dglemos: Which one? The one above doesn't have a dataset response and without a dataset, there's no url.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
```diff
@@ -541,61 +553,77 @@ sub variant_exists {
...
+      # convert to SO term
+      # There's more than one term for each type - use a list of SO terms
+      # CNV =~ DEL or DUP 
       my %terms = (
-        INS  => 'insertion',
-        DEL  => 'deletion',
-        CNV  => 'copy_number_variation',
-        DUP  => 'duplication',
-        INV  => 'inversion',
+        'INS'    => 'insertion',
```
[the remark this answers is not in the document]
dglemos: That's a good suggestion. This is one of the changes I need to make but I would prefer to do it in a different PR. I also need to change the `variant_exists` - it's too big and difficult to read at this point. I was thinking about changing everything after.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
dglemos: regex deleted - using `contains_value` seems like a better idea.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
dglemos: Thanks for the suggestions.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
dglemos: `contains_value` used instead.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
dglemos: GA4GH doesn't expect multiple URLs, so I didn't know where to put the different URLs. Merging them seemed ok but I'm also unsure what's the best approach.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
dglemos: I missed it. Thanks!

t/gabeacon.t
nerdstrike: Ideally you would mock up some test data that would work that part of the code and show the effect of none/one/many URLs. I appreciate that it's a royal pain.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
nerdstrike: Tell them their spec is broken for real data. You shouldn't violate the schema as it stands, so the workaround is the only option at this time.

lib/EnsEMBL/REST/Model/ga4gh/Beacon.pm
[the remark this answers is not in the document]
nerdstrike: Understood

PR #411, opened 2020-03-03
Task101/vep add more mf attribs

at7:
### Requirements
There is a dependency on merging the PR to ensembl-vep first: https://github.com/Ensembl/ensembl-vep/pull/703 The PR 703 will add new key value pairs to the VEP output which are tested for in ensembl-rest/vep.t

### Description
We are adding more keys to the VEP REST endpoint output. We are adding transcription_factors to a motif_feature_consequences array. So far the tests in vep.t only checked return values for the transcript_consequences array. We are adding tests for motif_feature_consequences and regulatory_feature_consequences arrays.

### Use case
Use VEP endpoint to annotate any overlaps with transcripts, regulatory or motif feature. For each overlap we calculate the variant consequence and depending on the feature add additional values.

### Benefits
Adding transcription_factors information provides more detailed description of the variant motif_feature overlap.

### Possible Drawbacks
Users might not be aware that we are returning one new key value pair which can potentially break parsers.

### Testing
Added tests to vep.t to test the expected structure of VEP endpoints.

### Changelog
[/vep/*] VEP endpoints updated to contain transcription_factors key for motif_feature_consequences.


at7:
https://github.com/Ensembl/ensembl-vep/pull/703 has been merged.

at7:
Tests (phenotype_region.t, variation.t, phenotype_gene.t) fail because of a new variation database patch for release 101 which hasn't been applied to the variation rest test database.

PR #412, opened 2020-03-06
GA4GH Beacon - clean code 

dglemos:
### Description

Clean some methods in the Beacon code. This PR is only to clean the code and not change any functionally. 

### Use case

`variant_exists` sub was too big and some variables usage could be improved. 

### Benefits

Make the code easier to read and understand. 

### Possible Drawbacks

None

### Testing

Yes, one test was added to cover one example not tested before. 

All the tests pass. 

Yes, no regression detected. 

### Changelog

No


ens-bwalts:
Would like to see this pass variation tests (which means we need variation test db patches) before approving

PR #421, opened 2020-03-26
Phenotype endpoint updates -e101

ima23:
### Description

Adding 5 more option parameters to allow fetch of phenotype annotations subsets:  `trait`, `tumour`, `non_specified`.

### Use case

_Describe the problem. Please provide an example representing the motivation behind the need for having these changes in place._

Currently all phenotype annotations are returned, however only some could be of interest. With these options only specific subsets of phenotypes will be retrieved. This reflects updates in how we store the data. 

### Benefits

Will be able to filter non specified phenotype annotations from the fetch result.

### Possible Drawbacks

None.

### Testing

_Have you added/modified unit tests to test the changes?_
Yes, I have added tests to cover the updates and updated the corresponding test data.

_If so, do the tests pass/fail?_
Yes, they pass.

_Have you run the entire test suite and no regression was detected?_
Yes, no regression detected.
Docs: http://hx-noah-12-02:3000/documentation/info/phenotype_gene
Tested [updated]: http://hx-noah-12-02:3000/phenotype/gene/homo_sapiens/LRRK2?content-type=application/json;include_associated=1;trait=1

### Changelog

_Are you changing the functionality of an endpoint? If so, please give a one line summary for the public facing changelog._

[/phenotype/gene/:species/:gene] new options added: trait, tumour, non_specified
[/phenotype/region/:species/:gene] new options added: trait, tumour, non_specified

PR #432, opened 2020-06-11
update parameter definition -e101

ima23:
### Description

In e101 phenotype endpoint has been extended by a few parameters. The parameter definition from 2 of them is here being updated to be clearer (the functionality was already added for e101 in PR https://github.com/Ensembl/ensembl-rest/pull/421).

### Use case

Users will understand better the effect of the new parameters.

### Benefits

Improved clarity.

### Possible Drawbacks

None.

### Testing

Have tested visually the display of the docs.

### Changelog

NA

### DIFF
--- a/root/documentation/phenotypefeatures.conf
+++ b/root/documentation/phenotypefeatures.conf
@@ -226,13 +226,13 @@
       </include_submitter>
       <trait>
         type=Boolean(0,1)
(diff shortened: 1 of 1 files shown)


ima23:
Tagging also @sarahhunt as in our SOP.

PR #452, opened 2020-09-02
update tests for new variant recoder return structure

at7:
### Requirements
We changed the output structure of the variant recoder tool. The code update is in ensembl-vep.

### Description
Variant recoder takes a given variant identifier as input and translates it to any other identifier which also describes the same variant. Some of the identifiers are variant allele specific. Currently all identifiers are returned together even if they refer to different alleles of the input variant. With the update we report identifiers by variant allele.

### Use case
The update helps users to only select identifiers for the allele they are interested in.
We are also planning a new web tool for which we want to report results by allele.

### Benefits
It is cleaner to report results by allele.

### Possible Drawbacks
Users will have written parsers for the current format and need to update their code to match the new structure.

### Testing
I have modified tests to comply with new structure in variation.t

Tests pass. Running entire test suite passes too https://travis-ci.org/github/at7/ensembl-rest/builds/723487977

### Changelog

We are changing the structure of the return value for the variant recoder endpoints:
GET variant_recoder/:species/:id
POST variant_recoder/:species
Identifiers are now reported separately for each alternative allele of the input variant.

### DIFF
--- a/t/variation.t
+++ b/t/variation.t
@@ -265,40 +265,42 @@ $base = '/variant_recoder/homo_sapiens';
 
 my $exp = [
   {
-    'hgvsp' => [
-      'ENSP00000371073.2:p.Asp22Glu',
-      'ENSP00000371079.3:p.Asp22Glu',
-      'ENSP00000381976.1:p.Asp22Glu',
-      'ENSP00000399510.1:p.Asp22Glu',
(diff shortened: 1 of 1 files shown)


ens-bwalts (approved):
Looks like all the test cases are updated to check the new by-allele structure

ameya1981 (approved):
Looking fine. tests are passing.

PR #453, opened 2020-09-17
Updates/submodule update 102

ens-bwalts:
### Requirements

### Description

Updates the wiki submodule to match the previous release (101), and updates the version to 13.1

### Use case

We need to keep the wiki submodule synced up with the version, and the REST version needs to be incremented to indicate a non-breaking change in functionality

### Benefits

See use case

### Possible Drawbacks

The wiki will still need to be updated to reflect release/102 functionality when it goes live. We may go back to having the wiki reflect the most recent release on GitHub, not the release available on the live server, and just inform outreach so that they can answer user questions appropriately

### Testing

N/A, documentation change only

N/A, documentation change only

### Changelog

N/A

### DIFF
--- a/lib/EnsEMBL/REST.pm
+++ b/lib/EnsEMBL/REST.pm
@@ -64,7 +64,7 @@ use Catalyst qw/
 /;
 
 
-our $VERSION = '13.0';
+our $VERSION = '13.1';
 
 # Configure the application.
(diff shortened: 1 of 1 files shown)


ameya1981 (approved):
Approved for 102.

PR #457, opened 2020-09-23
Beacon ID update (102) 

dglemos:
### Description

Update the Ensembl Beacon ID

### Use case

The current Beacon ID is not compatible with the Network specifications. 

### Benefits

To be able to register the Beacon in the Elixir Beacon Network. 

### Possible Drawbacks

None

### Testing

The test was updated. 

Test pass

No regression detected.

### Changelog

No change in functionality.


ameya1981:
As discussed with @dglemos, The link https://grch37.ensembl.org/Homo_sapiens/StructuralVariation/Explore?sv=esv93078 mentioned in the test only needs to be there for value check and does need to return a result

ameya1981 (approved):
Approved as tests are passing.

PR #458, opened 2020-09-23
Beacon ID update (103) 

dglemos:
### Description

Update the Ensembl Beacon ID. Same as #457 

### Use case

The current Beacon ID is not compatible with the Network specifications. 

### Benefits

To be able to register the Beacon in the Elixir Beacon Network. 

### Possible Drawbacks

None

### Testing

The test was updated. 

Test pass

No regression detected.

### Changelog

No change in functionality.


ameya1981 (approved):
Same as #457 . Approved.

PR #466, opened 2020-11-13
Add Starman to the cpanfile

tuanebi:
### Description

The same as https://github.com/Ensembl/ensembl-rest/pull/463, but to release/103


ameya1981 (approved):
Same as #463

PR #467, opened 2020-11-14
REeplace deprecated method with the correct one

mira13:
### Description

Removed getTotalLength method from assembly info, as it is deprecated and replaced with getRefLength

### Use case

After deprecated methods removal, build fails as we use one deprecated method in Assembly.pm

### Benefits

Correct endpoint work

### Possible Drawbacks

Not found

### Testing

No

### Changelog

endpoint assembly_info doesn't have "base_pairs" parameter any more.


ameya1981 (approved):
Reflects the removal of deprecated method as in https://github.com/Ensembl/ensembl/pull/534/files

magaliruffier:
Changes look fine, field should have been removed a while ago

magaliruffier (approved):

magaliruffier:
Travis failures are on variation endpoints and are not related to this change.
Has the Variation team been informed?

PR #471, opened 2020-12-15
Task104/variation test db patch

at7:
### Description

Patch variation test database for 104.

### Use case
NA

### Benefits
NA

### Possible Drawbacks
NA

### Testing
I had to switch off one test which was failing because of a problem with the conversion to hex-blob https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-3630

Tests pass https://travis-ci.org/github/at7/ensembl-rest/builds/749818960

### Changelog
No


ameya1981 (approved):
version bump changes match those in variation repo for 104

PR #473, opened 2020-12-15
Change test db blob to hex format

mira13:
### Description

Replace hex db blob with hex format

### Use case

We are no more using binary format, so need to replace it

### Benefits

Tests will pass correctly

### Possible Drawbacks

Not found

### Testing

No, this is to fix build

yes, they passes

### Changelog

Nothing, just to fix build


ameya1981 (approved):
Approved for 104. Variation will have to update their repo with this format

PR #476, opened 2021-01-14
add test back after blob to hex-blob conversion

at7:
### Description

Variation test data has been converted to HEX and the test runs successfully again.

### Use case

NA

### Benefits

NA

### Possible Drawbacks

### Testing

Tests ran on travis https://travis-ci.org/github/at7/ensembl-rest/builds/754441684

### Changelog

NA


ens-bwalts (approved):
Underlying change in test data was merged in December (I believe this was before `release/104` branching). This is just a test, so I don't see a need to try and bring this into `release/104`, but would be nice to get it in before 105 branching.

magaliruffier (approved):
the test already existed and now it's even being run, what's not to like :-)

PR #477, opened 2021-02-15
Patching test DBs for 105 version bump

TamaraNaboulsi:
### Description

Changes from running patch_test_databases.pl on this repo. In addition, fixed meta.txt file for homo_sapiens/funcgen which contained spaces instead of tabs for last 2 patches added, which was causing wrong patching output.


mira13 (approved):
Version and schema update for 105, looks fine

PR #487, opened 2021-05-13
Add options to Variant Recoder

dglemos:
### Description

Variant Recoder has two more options available. 

### Use case

We need to add two options:
- vcf_string (at the moment, it can be switched on using `fields` but it's more correct if `vcf_string` is a separate option)
- var_synonyms (new option to report variation synonyms)

### Benefits

More Variant Recoder options.

### Possible Drawbacks

None

### Testing

Yes, tests added to t/variation.t

Tests pass

No regression detected.

### Changelog

[variant_recoder/:species/:id] Has the option to return VCF and variation synonyms

### DIFF
--- a/root/documentation/variation.conf
+++ b/root/documentation/variation.conf
@@ -176,9 +176,19 @@
       </species>
       <fields>
         type=String
-        description=Comma-separated list of identifiers/notations to include from the following types: id (variant ID), hgvsg (HGVS genomic), hgvsc (HGVS coding), hgvsp (HGVS protein), spdi (SPDI genomic), vcf_string (VCF format)
+        description=Comma-separated list of identifiers/notations to include from the following types: id (variant ID), hgvsg (HGVS genomic), hgvsc (HGVS coding), hgvsp (HGVS protein), spdi (SPDI genomic)
         default=id,hgvsg,hgvsc,hgvsp,spdi
       </fields>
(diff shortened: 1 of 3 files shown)


root/documentation/variation.conf
```diff
@@ -176,9 +176,19 @@
       </species>
       <fields>
         type=String
-        description=Comma-separated list of identifiers/notations to include from the following types: id (variant ID), hgvsg (HGVS genomic), hgvsc (HGVS coding), hgvsp (HGVS protein), spdi (SPDI genomic), vcf_string (VCF format)
+        description=Comma-separated list of identifiers/notations to include from the following types: id (variant ID), hgvsg (HGVS genomic), hgvsc (HGVS coding), hgvsp (HGVS protein), spdi (SPDI genomic)
         default=id,hgvsg,hgvsc,hgvsp,spdi
       </fields>
+      <vcf_string>
+        type=String
```
ens-bwalts, line 183: Having looked at how these options are used in the test case - would it be better to describe the type of these options as `Boolean(0,1)` ?

dglemos: It makes more sense to be a boolean, thanks for pointing that out.

ens-bwalts (approved):
Syntax checks, and test cases added which are passing.

PR #489, opened 2021-06-07
Feature/mane update

magaliruffier:
### Requirements

### Description

Add MANE data to lookup and overlap endpoints. The data is only available with an additional option, to avoid modifying the output of existing endpoints.
Also add is_canonical for transcript and canonical_transcript attributes for gene where not already available.

### Use case

MANE transcripts are transcripts that are identical and agreed between Ensembl and RefSeq.
Users want to know:
- if a transcript is part of MANE
- what RefSeq accession the Ensembl transcript maps to

### Benefits

MANE data can be retrieved through the REST API

### Possible Drawbacks

MANE data is not always straightforward to retrieve.
This is to avoid modifying behaviour of existing endpoints, advanced users who are interested in MANE data will need additional options.

### Testing

Yes

Pass

No regression noticed, although some tests were failing before the changes.

### Changelog

*Updated endpoints*
[/lookup] and [/overlap] Added the canonical_transcript attribute to gene object, to provide the stable ID of the canonical transcript for that gene
[/lookup] Added MANE object. Available only with the expand=1 and mane=1 option
[/overlap] Added MANE object. Available with the feature=mane option.

*Bugfixes*
[/overlap] Added is_canonical attribute for transcript feature, for consistency with the lookup endpoint
[/lookup] Added version attribute for translation feature in full format (already available in condensed format)


root/documentation/lookup.conf
```diff
@@ -40,6 +40,11 @@
         description=Include 5' and 3' UTR features. Only available if the expand option is used.
         default=0
       </utr>
+      <mane>
```
ens-bwalts, line 43: Does this need to be included under the POST documentation as well?

ens-bwalts (approved):
There are a few things packed into this PR. Going through them:
First, the update to GFFSerializer, which is necessary and might as well go here

Second, the updates to the Overlap endpoints. These now look for MANE overlaps if requested, and has method to handle MANE objects. It also now finds canonical transcripts and adds annotations to genes and transcripts. These all have tests, and the documentation is updated.

Third, there are the updates to the lookup endpoints to handle MANE transcripts. This change set updates feature_as_hash to handle versions, canonical transcripts, as well as MANEs and their refseq matches. It also provides a MANE method to handle MANE objects. Tests are in place and passing. The documentation is updated as well - I think I answered my own question and the POST request does not take a mane Boolean, but I'll leave the question in place for verification.

One potential trouble point is the inconsistent capitalisation of 'MANE', which is sometimes all-caps (hash keys, sub MANE in Lookup.pm, and sometimes all lower-case (object type, sub mane in Overlap). Unfortunately, these are sometimes due to technical necessity (e.g. the capitalisation of the two different sub mane-s). In this PR, it looks like the capitalisation is correct for the particular situation throughout.

TamaraNaboulsi (approved):
- Added methods to handle MANE data in both overlap and lookup endpoints
- MANE method only called when requested with additional options
- Added tests and updated documentation in both endpoints
- Fixed deprecated GFFSerializer call
- Added is_canonical attribute for transcripts and canonical_transcript attribute for gene where needed

All checks are passing and changes look good.

PR #491, opened 2021-06-09
Patch variation test db 105 

dglemos:
### Description

Variation has another schema patch for 105. 

### Use case

Variation needs to add another schema patch to 105 because the publication.title current length is not enough. 

### Possible Drawbacks

None

### Testing

No

### Changelog

No

### DIFF
--- a/t/test-genome-DBs/homo_sapiens/variation/table.sql
+++ b/t/test-genome-DBs/homo_sapiens/variation/table.sql
@@ -186,7 +186,7 @@ CREATE TABLE `meta` (
   PRIMARY KEY (`meta_id`),
   UNIQUE KEY `species_key_value_idx` (`species_id`,`meta_key`,`meta_value`),
   KEY `species_value_idx` (`species_id`,`meta_value`)
-) ENGINE=MyISAM AUTO_INCREMENT=132 DEFAULT CHARSET=latin1;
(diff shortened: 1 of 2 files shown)


TamaraNaboulsi (approved):
The size of 'title' column of table 'publication' was increased to 400 characters. Looks good.

at7 (approved):

t/test-genome-DBs/homo_sapiens/variation/table.sql
```diff
@@ -325,7 +325,7 @@ CREATE TABLE `protein_function_predictions_attrib` (
 
 CREATE TABLE `publication` (
   `publication_id` int(10) unsigned NOT NULL AUTO_INCREMENT,
-  `title` varchar(255) DEFAULT NULL,
+  `title` varchar(400) DEFAULT NULL,
```
ens-bwalts, line 328: @at7 @dglemos In the variation schema, this is a varchar(300), not 400. https://github.com/Ensembl/ensembl-variation/blob/be5c085fbb8f829562b866d1414b6fe5d9f949b5/sql/table.sql#L1741

Don't think it matters for a test DB patch, but wanted to point out the discrepancy in case it indicates an issue with the patcher.

dglemos: @ens-bwalts the PR for this patch wasn't merged yet. We are waiting for final confirmation from production.

PR #502, opened 2021-10-08
Update spliceai input file

dglemos:
### Description

Same as https://github.com/Ensembl/ensembl-rest/pull/501 
Add option to SpliceAI plugin: allow users to select a different snv input file.

### Use case

Currently, the annotation scores are only retrieved from the Spliceai Illumina file. We want users to have the option to retrieve scores from the Ensembl MANE transcripts annotation file (SNV only).

### Benefits

The Ensembl MANE transcript scores will be available to users.

### Possible Drawbacks

None

### Testing

No

No regression detected

### Changelog

Added the option to select SpliceAI annotation scores from Ensembl MANE transcripts (SNV only)
[vep/:species/hgvs/:hgvs_notation]
[vep/:species/id/:id]
[vep/:species/region/:region/:allele/]


ens-bwalts (approved):
Same as #501

PR #512, opened 2021-11-22
Increase gene_tree_root.clusterset_id column length

JAlvarezJarreta:
### Description

Compara has included wheat cultivars protein trees in e106 and this has flagged a requirement to increase the column of one of our tables.

### Testing

This patch has been successfully applied to all Compara production pipelines and databases in e106 without raising any incompatibilities or issues.

### Changelog

_Are you changing the functionality of an endpoint? If so, please give a one line summary for the public facing changelog._
No

PR #513, opened 2021-12-07
Increase gene_tree_root.clusterset_id column size

sgiorgetti:
### Description

Compara has included wheat cultivars protein trees in e106 and this has flagged a requirement to increase the column of one of our tables.

It is the same of PR 512 to be merged against "master".

### Testing

On Codon, two tests (2 and 6) from gafeatures.t fail because the value for "updated" and "created" features attributes "2008-04-29 11:17:41" is converted to epoch "1209464261" instead of the expected "1209467861".
We are not changing test data because of Travis automatic build/test.

Also on Codon, there are ld.t tests failing because of ld_vcf (in ensembl-variation repo) segfaulting. This is an issue resolved by https://github.com/Ensembl/ensembl-variation/pull/770

### Changelog

No


TamaraNaboulsi (approved):
Same as #512 but for master.
clusterset_id column size increased in table gene_tree_root and patch added accordingly

PR #518, opened 2022-01-26
Patching variation test db with new consequences

jamie-m-a:
Three additional variant consequence are being added to the variation test DB


jamie-m-a:
Tagging @dglemos as per SOP

dglemos (approved):

ens-bwalts (approved):
Matches variation's 106_107_b

PR #523, opened 2022-02-14
Patch test-dbs for compara case insensitive stable_id schema alteration

CristiGuijarro:
### Description

_In e107 Metazoa compara revealed an issue in that the schema expected entirely `UNIQUE` `stable_id` across all genomes. This cannot be the case because divisions such as Metazoa have external annotations and `UNIQUE` `stable_id` is not enforced, not expected._

### Use case

_The compara schema and API is too tied in with a `UNIQUE` enforcement and may require a larger change later on. For now, changing the case sensitivity is a passable compromise to get compara through e107. (The conflicting `stable-ids` remain unique when case is taken into account)._

### Benefits

_This small change allows the compara schema and API to remain largely unchanged, whilst allowing through the current e107 dataset._

### Possible Drawbacks

_There are no consequences, bar the fact that future conflicts may require a larger and more complex change. This is enough for now._

### Testing

_Have you added/modified unit tests to test the changes?_ NA

_If so, do the tests pass/fail?_ NA

_Have you run the entire test suite and no regression was detected?_  :heavy_check_mark:

### Changelog

NA

### DIFF
--- a/t/test-genome-DBs/homology/compara/meta.txt
+++ b/t/test-genome-DBs/homology/compara/meta.txt
@@ -77,3 +77,4 @@
 103	\N	patch	patch_105_106_a.sql|schema_version
 105	\N	patch	patch_106_107_a.sql|schema_version
 106	\N	patch	patch_105_106_b.sql|clusterset_id_varchar50
+107	\N	patch	patch_106_107_b.sql|case_sensitive_stable_id
--- a/t/test-genome-DBs/homology/compara/table.sql
+++ b/t/test-genome-DBs/homology/compara/table.sql
@@ -138,7 +138,7 @@ CREATE TABLE `gene_align_member` (
 
 CREATE TABLE `gene_member` (
   `gene_member_id` int(10) unsigned NOT NULL AUTO_INCREMENT,
-  `stable_id` varchar(128) NOT NULL,
+  `stable_id` varchar(128) CHARACTER SET latin1 COLLATE latin1_bin NOT NULL,
   `version` int(10) unsigned DEFAULT '0',
   `source_name` enum('ENSEMBLGENE','EXTERNALGENE') NOT NULL,
   `taxon_id` int(10) unsigned NOT NULL,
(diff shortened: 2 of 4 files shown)


sgiorgetti (commented):
Changes are limited to the Compara DB.
meta contents look good.
As per the use of 'latin1_bin', not sure why we want to compare stable_id's as per binary rules, instead of character ones?

CristiGuijarro:
> Changes are limited to the Compara DB. meta contents look good. As per the use of 'latin1_bin', not sure why we want to compare stable_id's as per binary rules, instead of character ones?

This is so that the `stable_id` can be compared byte-by-byte rather than character-by-character - otherwise the `Stable_id` will be the same as `stable_id` when compared for `UNIQUE` -ness

sgiorgetti:
@CristiGuijarro - I understand you are seeking for a case-sensitive comparison between strings (stable IDs), but my two points are
1. Why not using `latin1_general_cs` instead `latin1_bin`, as it's meant to work for strings, rather than binary info.
Also, the use of binary collation for strings might have some unexpected impact on sorting. Not sure it's important or applicable in your use case, though.
2. The collation is usually set and managed at DB level, and people are expecting this. It is true you may attach a collation at table level and column level, but this is uncommon. Also, this bit of information gets "hidden" in the DDL for table definition, hence not immediately visible. My concern here is about the maintainability of your DB in the future.

This said, it's Compara DB schema we are talking about, and Compara are the owners of this: if you are sure and want to go on, please just confirm so and I'll be happy to go ahead and merge your PR.

CristiGuijarro:
> @CristiGuijarro - I understand you are seeking for a case-sensitive comparison between strings (stable IDs), but my two points are
> 
> 1. Why not using `latin1_general_cs` instead `latin1_bin`, as it's meant to work for strings, rather than binary info.
>    Also, the use of binary collation for strings might have some unexpected impact on sorting. Not sure it's important or applicable in your use case, though.
> 2. The collation is usually set and managed at DB level, and people are expecting this. It is true you may attach a collation at table level and column level, but this is uncommon. Also, this bit of information gets "hidden" in the DDL for table definition, hence not immediately visible. My concern here is about the maintainability of your DB in the future.
> 
> This said, it's Compara DB schema we are talking about, and Compara are the owners of this: if you are sure and want to go on, please just confirm so and I'll be happy to go ahead and merge your PR.

Thanks @sgiorgetti - this is a stop-gap and not a long term solution, we have a ticket for a long-term solution, the schema change is already in the e107 compara release, so please go ahead and merge :+1:

sgiorgetti (approved):
Approved with reservation.
Please, see conversation history.

PR #524, opened 2022-02-17
feat: Add VEP Plugin - EVE

diegomscoelho:
### Description

The VEP endpoint will have the external plugin EVE available. This documentation is needed for the plugin.

### Use case

### Benefits

### Possible Drawbacks

### Testing

Documentation check only.

### Changelog

[/vep/:species/hgvs/] new plugin option added: EVE
[/vep/:species/id/] new plugin option added: EVE
[/vep/:species/region/] new plugin option added: EVE


diegomscoelho:
Just to advise, @nakib103 already tested this feature and said it's working!

PR #525, opened 2022-03-30
Add EVE plugin

sgiorgetti:
### Description

The VEP endpoint will have the external plugin EVE available. This documentation is needed for the plugin.
Same as PR #524

### Use case

### Benefits

### Possible Drawbacks

### Testing

Documentation check only.

### Changelog

[/vep/:species/hgvs/] new plugin option added: EVE
[/vep/:species/id/] new plugin option added: EVE
[/vep/:species/region/] new plugin option added: EVE


TamaraNaboulsi (approved):
Same as #524 but for release/107. Plugin EVE added to id, region, and hgvs (get and post). Looks good.

PR #527, opened 2022-04-04
Updated change-log for EVE plugin

TamaraNaboulsi:
### Description

Updating the docs (wiki) change-log to include EVE plugin.

### Use case

EVE plugin needed to be added into 107, but that had already branched, so the docs submodule needed to be updated.

### Benefits

Updated wiki

### Possible Drawbacks

None

### Testing

### Changelog


sgiorgetti (approved):
Fine by me

PR #530, opened 2022-04-14
New VEP plugin GO

nuno-agostinho:
### Description

The VEP endpoint will have the external plugin GO available. This documentation is needed for the plugin.

### Use case

### Benefits

### Possible Drawbacks

### Testing

Documentation check only.

### Changelog

[/vep/:species/hgvs/] new plugin option added: GO
[/vep/:species/id/] new plugin option added: GO
[/vep/:species/region/] new plugin option added: GO


TamaraNaboulsi (approved):
Added GO plugin to VEP id, region, and hgvs (get and post). Looks good.
I will be adding similar PR against release/107 since that has branched already.

PR #531, opened 2022-04-14
Add docs for VEP GO plugin for 107

TamaraNaboulsi:
Same as #530 but against release/107


sgiorgetti (approved):
Same as #530 
Looks good

PR #517, opened 2022-01-19
New VEP plugin UTRannotator (e107)

dglemos:
### Description

The VEP endpoint will have the external plugin UTRannotator available. This documentation is needed for the plugin.

### Use case

Plugin functionality available through REST.

### Benefits

The endpoint needs documentation to be displayed.

### Possible Drawbacks

None, the default behavior is the same.

### Testing

note to submitters and reviewers: documentation-only changes may reflect
changes in other repos that can result in new or different output from
REST endpoints. In turn, these may require new tests or changes to existing tests.

_Have you added/modified unit tests to test the changes?_

_If so, do the tests pass/fail?_

_Have you run the entire test suite and no regression was detected?_
No regression was detected.

### Changelog

_Are you changing the functionality of an endpoint? If so, please give a one line summary for the public facing changelog._

eg. [/xenobiology/orthologs] Added the ability to look up orthologs and paralogs from klingons and andorians
[/vep/:species/hgvs/] new plugin option added: UTRannotator
[/vep/:species/id/] new plugin option added: UTRannotator
[/vep/:species/region/] new plugin option added: UTRannotator

PR #538, opened 2022-06-01
Remove VEP plugin UTRannotator (108)

dglemos:
### Description

The VEP endpoint will not have the external plugin UTRannotator available.
This PR removes docs added in https://github.com/Ensembl/ensembl-rest/pull/517

### Use case

Plugin functionality won't be available through REST.

### Possible Drawbacks

None

### Testing

### Changelog

Revert the functionality of:
[/vep/:species/hgvs/]
[/vep/:species/id/]
[/vep/:species/region/]


TamaraNaboulsi (approved):
Same as #537 but for the main branch (since this has branched to 107 already). The travis error is occurring because the ensembl repo has already moved on to 108 (this PR wouldn't normally happen this late in the process). Look good.

PR #542, opened 2022-06-28
Replace deprecated `MemberAdaptor::fetch_by_stable_id`

CristiGuijarro:
### Requirements

### Description

In e109 compara are deprecating the method `MemberAdaptor::fetch_by_stable_id()` - this PR addresses the replacement of this method with new method `MemberAdaptor::fetch_by_stable_id_GenomeDB()`.
See ticket: [ENSCOMPARASW-5270](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-5270).
Related PR: https://github.com/Ensembl/ensembl-compara/pull/469

### Use case

A `stable_id` is only unique within a genome and not across a compara database. This has always been a case, but until recently, and purely by chance, there have been no overlapping `stable_id`. As the number of species included in EnsEMBL increases this becomes a big problem, and so the change is urgent-ish and necessary.

### Benefits

The benefits include not breaking `ensembl-compara` or `ensembl-rest` and the ability to include all/any combination of species within compara, regardless of duplicated `stable_ids` in genes/proteins etc.
 
### Possible Drawbacks

This limits some of the flexibility of being able to search for some homologies for example by `stable_id` alone - as can be seen in [lib/EnsEMBL/REST/Controller/Homology.pm](https://github.com/Ensembl/ensembl-rest/compare/main...CristiGuijarro:ensembl-rest:fix/compara_stable_ids?expand=1#diff-ef99ce26b7e6fc955d2d8fc06f302f82c5108f60bdb1142703c35ebdfce6ca48).

### Testing

No

The individual tests passed locally - will see how it goes here.

No regression was detected between changed and unchanged - however it was noticed in local testing that there is/was an issue Variation side. Will find out here.

### Changelog

Nope. Although I do think we should remove the functionality to search for homologs using just a gene `stable_id` and no `species` - have a fix in place for this to simply select the first `gene_member` with the `stable_id` - but this method makes no sense now or before when `stable_ids` are only unique within a genome and not outside of a genome. As it stands pre and post this PR - a user can search for a homolog with a specific gene in mind and come up with a gene from an unexpected species with the same `stable_id` instead - this is particularly the case for non-vert and community submitted genomes in which gene `stable_ids` have been named based on homology in the first place. This change would be outside of the scope of this PR.


sgiorgetti (approved):
Looks good to me.

PR #545, opened 2022-08-11
New VEP plugin mutfunc (e108)

nakib103:
### Description

We are adding new plugin for VEP. This PR add the documentation for that.

### Use case

Plugin functionality available through REST.

### Benefits

The endpoint needs documentation to be displayed.

### Possible Drawbacks

None, the default behavior is the same.
Should only be merged if https://github.com/Ensembl/ensembl-rest_private/pull/119 is merged

### Testing

Tested on sanbox.

### Changelog

[/vep/:species/hgvs/] new plugin option added: mutfunc
[/vep/:species/id/] new plugin option added: mutfunc
[/vep/:species/region/] new plugin option added: mutfunc

PR #546, opened 2022-08-27
New VEP plugin mutfunc

TamaraNaboulsi:
### Description

Same as #545 against `main` branch

NOTE: Once this is approved and merged, the Wiki needs to be updated to include the new changelog (lines added into JIRA ticket


sgiorgetti (approved):
Same as #545 - `mutfunc` added to VEP plugins
Look good

PR #555, opened 2022-09-14
Update Beacon to include dbSNP as dataset

dglemos:
### Description

Beacon reports if variants are found in different datasets. Our implementation of Beacon is not considering dbSNP as a dataset which is not correct. This PR adds dbSNP as a dataset and reports accordingly. 

### Use case

Beacon is not using dbSNP as a dataset which can be a problem: a few queries report that a variant exists in Ensembl but they have no attached dataset in the response. This happens because these variants were only reported in dbSNP. 
Example: http://rest.ensembl.org/ga4gh/beacon/query?content-type=application/json;referenceName=1;referenceBases=G;start=234010095;assemblyId=GRCh38;alternateBases=T;includeDatasetResponses=HIT
The `datasetAlleleResponses` is empty but it should include the dbSNP dataset.

### Benefits

Fix a bug when including the dataset response.

### Possible Drawbacks

None

### Testing

Yes, I added a test.

Tests pass

None detected

### Changelog

[/ga4gh/beacon] Added the dataset dbSNP
[/ga4gh/beacon/query] Fix bug: the response includes the dataset dbSNP if the query includes the dataset response


nakib103 (changes requested):

nakib103 (approved):
Tested in sandbox. Working as intended. 
LOTM! ~

sgiorgetti (approved):
Changes to the beacon endpoint and related test case.
Look OK to me and are approved by Variation.

PR #579, opened 2023-01-19
main: Replace probe related to the retired microarray in Regulatory documentation pages

pblins:
Description

This PR is to replace probe related to the retired microarray with a valid one ( ILMN_1763508 -> ILMN_1910180 ).


sgiorgetti (approved):
Looks good

PR #580, opened 2023-01-23
Revert fetch_by_ensembl_gene to release/108 version

twalsh-ebi:
### Description

This change reverts method `fetch_by_ensembl_gene` to its `release/108` version, and in doing that restores the ability to query the `homology/id/:id` endpoint by versioned gene stable ID.

### Use case

Until Ensembl 108, queries of the `homology/id` REST endpoint would return a result for a versioned gene stable ID, even if the initial stable_id_lookup failed, because of a fallback query of Compara databases to verify that the gene could indeed be accessed by the given versioned gene stable ID. As part of the process of deprecating Compara method `MemberAdaptor::fetch_by_stable_id`, that fallback query was removed, which effectively removed support for querying homologies by versioned gene stable ID.

For example, these queries return equivalent results via the Ensembl 108 REST API:
- https://rest.ensembl.org/homology/id/ENSG00000186092?content-type=application/json;compara=vertebrates;target_species=equus_caballus
- https://rest.ensembl.org/homology/id/ENSG00000186092.7?content-type=application/json;compara=vertebrates;target_species=equus_caballus

However, the latter query currently fails on the Ensembl 109 test REST site.

With the extension of the deprecation period of `MemberAdaptor::fetch_by_stable_id`, it is possible to reinstate the fallback query and thereby restore the ability to query homologies by versioned gene stable ID.

### Benefits

Users will be able to continue using the `homology/id/:id` endpoint to query homologies by versioned gene stable ID for at least one more Ensembl release. Development is in progress to replace this functionality in release 110.

### Possible Drawbacks

This solution is necessarily temporary, since `MemberAdaptor::fetch_by_stable_id` will be removed from the Compara Perl API in Ensembl 110. But in the worst case, support for querying homologies by versioned gene stable ID would end in Ensembl 110 rather than release 109.

### Testing

A unit test has been added to `homology.t` which queries the `homology/id/:id` endpoint with horse gene stable ID `ENSECAG00000014890`. This is not a versioned stable ID, but it can serve as a proxy for one, because it is present in the `homology` Compara unit-test database but not in the `stable_ids` test database used in stable_id_lookup.

All Compara-related unit tests passed locally.

### Changelog

No change as such, only a partial reversion of the changes in PR #542 to restore functionality present in the Ensembl 108 REST API.

### DIFF
--- a/lib/EnsEMBL/REST/Controller/Homology.pm
+++ b/lib/EnsEMBL/REST/Controller/Homology.pm
@@ -72,12 +72,25 @@ sub fetch_by_ensembl_gene : Chained("/") PathPart("homology/id") Args(1)  {
   my ( $self, $c, $id ) = @_;
   my $lookup = $c->model('Lookup');
   my $species;
-  ($species) = $lookup->find_object_location($id);
+  try {
+    ($species) = $lookup->find_object_location($id);
+  } catch {
+    # if not found in stable_id_lookup for some reason (e.g. versioned gene stable ID), try looking it up in compara dbs
+    my $comparas = $c->model('Registry')->get_all_DBAdaptors('compara');
+    my $gene_member;
+    foreach my $compara_db ( @$comparas ) {
+      my $gma = $compara_db->get_GeneMemberAdaptor();
+      $gene_member = $gma->fetch_by_stable_id($id);
+      if ( $gene_member ) {
+        last;
+      }
+    }
 
-  if ( !$species ) {
-    $c->go('ReturnError', 'from_ensembl', [qq{$_}]) if $_ =~ /STACK/;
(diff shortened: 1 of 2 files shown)


sgiorgetti (approved):
Look good to me

sgiorgetti:
@twalsh-ebi - for the records these changes have been reverted from 'release/109', but they are still in 'main' and shall go into 'release/110' as soon as we branch.
This is the expected behaviour we agreed on. Thx

twalsh-ebi:
> @twalsh-ebi - for the records these changes have been reverted from 'release/109', but they are still in 'main' and shall go into 'release/110' as soon as we branch. This is the expected behaviour we agreed on. Thx

Thanks @sgiorgetti, I understand. An update to the reverted method is in development for the 'main' branch.

PR #599, opened 2023-07-03
Dev/lookup transcript length

sgiorgetti:
### Description

Adding transcript length to `Lookup` endpoint

### Use case

Have immediately available this bit of information.

### Benefits

See above

### Possible Drawbacks

None

### Testing

Tests were modified and could run successfully

### Changelog

N/A


nwillhoft (approved):
PR to add in transcript length to REST endpoint. The attribute has been added in for the lookup and a couple of tests have been added into `t/lookup.t`. Looking at the build logs, all of the tests in this file are passing. The builds are only failing because of a version mismatch (i.e. expecting 111 but getting 110), which I believe is not a problem as this will be resolved when we branch this repo to 111.

nwillhoft:
@shashankbrgowda @sgiorgetti with regards to adding this functionality to the REST change log. Would it be appropriate to add something like?:
`[/lookup/id/:id] Transcript object now returns transcript length`

Issue #588, opened 2023-04-11
GET overlap/ endpoints missing exons at edge of region


feliciazhang:
Hello! Running into a possible inclusive/exclusive region edge case with the `GET overlap/id/:id` and `GET overlap/region/:species/:region` endpoints. When fetching features from a gene ID/region where there is a exon at the start of the specified region that has a length of 1, that exon is not included in the response as expected.

For example:

Endpoint `overlap/id/Pf3D7_0807000` with parameters `feature=gene;feature=transcript;feature=cds;feature=exon` and content type `"text/x-gff3"` has the response below where the exon `CAX64101-1` with start and end at 370784 is missing. Per the exon table on the [Ensembl site](https://protists.ensembl.org/Plasmodium_falciparum/Transcript/Exons?db=core;g=PF3D7_0807000;r=8:370784-372623;t=CAX64101) I would expect this exon to be included in this region specified by the gene. The same result occurs when using the `/region/` version of the endpoint with the gene's region `overlap/region/plasmodium_falciparum/8:370784-372623`. Interestingly, CDS features at that same 370784-370784 position are included in the response when CDS is added to the feature parameters.

```
##gff-version 3
##sequence-region   8 370784 372623
8	ena	gene	370784	372623	.	+	.	ID=gene:PF3D7_0807000;assembly_name=ASM276v2;biotype=protein_coding;description=gas41 homologue%2C putative;gene_id=PF3D7_0807000;logic_name=ena
8	ena	mRNA	370784	372623	.	+	.	ID=transcript:CAX64101;Parent=gene:PF3D7_0807000;assembly_name=ASM276v2;biotype=protein_coding;logic_name=ena;transcript_id=CAX64101
8	ena	exon	370911	370987	.	+	.	Name=CAX64101-2;Parent=transcript:CAX64101;assembly_name=ASM276v2;constitutive=1;ensembl_end_phase=0;ensembl_phase=1;exon_id=CAX64101-2;rank=2
8	ena	exon	371150	371282	.	+	.	Name=CAX64101-3;Parent=transcript:CAX64101;assembly_name=ASM276v2;constitutive=1;ensembl_end_phase=1;ensembl_phase=0;exon_id=CAX64101-3;rank=3
[...]

PR #612, opened 2023-10-23
Hotfix/enscoresw 4269 e111

sgiorgetti:
### Description

Fix issue #588 

### Use case

Exons of length 1 at the boundary of the given region were not returned by the /overlap endpoint.

### Benefits

Return the exons of length 1 sitting at the region boundary. 

### Possible Drawbacks

Downstream pipelines can be affected.

### Testing

Test suite runs fine

### Changelog

N/A


TamaraNaboulsi (approved):
Updated the exon overlap logic to include the exons having a start be equal to the slice start and and end equal to the slice end (for exons not fully included in the slice itself). This specifically solves the edge case of exons with length 1 falling on the start or end coordinates of a slice.

All looks good. also had to patch travis yml for the test suite to work.

This will also need to be added into the main branch later on.

PR #613, opened 2023-10-23
Include exons of length 1 at the boundary

sgiorgetti:
### Description

Sister to PR #612 - this applied to main branch 
Fix issue #588 

### Use case

Exons of length 1 at the boundary of the given region were not returned by the /overlap endpoint.

### Benefits

Return the exons of length 1 sitting at the region boundary. 

### Possible Drawbacks

Downstream pipelines can be affected.

### Testing

Test suite runs fine

### Changelog

N/A


TamaraNaboulsi (approved):
Same as #612 against main branch.
Looks good

PR #634, opened 2024-03-19
Update test dbs to Release 113

nwillhoft:
Update the API version to 113 and add 113 db schema patch.

Use the 'b' patch (from [this PR](https://github.com/Ensembl/ensembl/pull/689)) to revert the meta_value to NOT NULL (previously changed in 112 upon request from Production).


t/test-genome-DBs/homo_sapiens/core/table.sql
```diff
@@ -486,11 +486,11 @@ CREATE TABLE `meta` (
   `meta_id` int(11) NOT NULL AUTO_INCREMENT,
   `species_id` int(10) unsigned DEFAULT '1',
   `meta_key` varchar(40) NOT NULL,
```
sgiorgetti, line 488: @nwillhoft  - the `meta_key` looks like always `VARCHAR(40)` whereas it should be `VARCHAR(64)

sgiorgetti (changes requested):
@nwillhoft  - it looks like the patcher did not pick up the `meta_key` extension to `VARCHAR(64)`. Not sure why. Mind take a look into it? Thx

t/test-genome-DBs/homo_sapiens/core/table.sql
nwillhoft: I believe all the `meta_key` values have now been updated

sgiorgetti (approved):
look good to me

PR #636, opened 2024-04-03
Allow REST to show presence of GENCODE Primary attribute

nwillhoft:
Allow REST to return boolean value for presence of GENCODE Primary attribute.

With GENCODE Primary being introduced for 112, do we also want to push this change onto the `release/112` branch?


nwillhoft:
Ready to go, along with https://github.com/Ensembl/ensembl/pull/696 and https://github.com/Ensembl/ensembl-io

sgiorgetti (approved):
OK

PR #637, opened 2024-04-15
Perl documentation fix

nwillhoft:
When printing out non-JSON content with perl, the example REST documentation is currently suggesting to use the response status and reason attributes. This PR updates the perl example to print out the response content instead.


sgiorgetti (approved):
ok

PR #640, opened 2024-04-25
Out-of-date reg. feature update (112)

pblins:
Description
Update out-of-date reg. feature in the documentation page .


nwillhoft (approved):
The example `regulation_id` is being updated in the REST documentation as the previous one is no longer used. The build checks are passing. Testing this change in my local REST server shows that the example id is updated accordingly.

PR #645, opened 2024-06-18
Update regulatory overlap test to reflect the most updated reg. summary [release/113]

pblins:
### Description

Update the regulatory feature summary response.

PR #649, opened 2024-06-20
Update regulatory overlap test to reflect the most updated reg. summary [release/113]

sgiorgetti:
### Description

Update the regulatory feature summary response.
Same as #645 - needed because of Travis issue.

Updated test as per #648 - this PR is for `release/113` branch, instead of `main`


nwillhoft (approved):
Changes look good and the branch build checks are passing. 

It appears this PR build check is failing because Travis cannot clone the `release/113` branch for the `ensembl-vep` repo (because it does not yet exist). Expected behaviour would then be to clone the `main` branch (i.e. defined in $SECONDARY_BRANCH), but it does not and so the repo cloning fails.

Approving this PR as it's unrelated to the Travis cloning issue.

PR #657, opened 2024-08-07
TravisCI update

sgiorgetti:
### Description

- Removing Perl 5.14, as EnsEMBL does not support it anymore
- Adding Perl 5.30 as "modern" version
- Bumping TravisCI build environment to "focal" - see also deprecation warning [HERE](https://docs.travis-ci.com/user/reference/linux/)
- fixed issue with cloning Perl API repos
- fixed issue with Coveralls integration
- bumped HTS lib version to 1.13
- fixed issue of loading Compara test database

### Use case

Perl 5.14 is not supported anymore, and it's been replaced by 5.26 as oldest supported version by EnsEMBL.
Also, TravisCI is going to deprecate Linux "trusty" build environment by EOY 2024.
Coveralls integration was not working, because of change in the API.

### Benefits

Enhanced efficiency and better use of resources, by focussing on supported Perl version.
Bug fixing.

### Possible Drawbacks

None

### Testing

Some Compara test data could not be loaded from file `t/test-genome-DBs/multi/compara/conservation_score.txt`.
Problem is that `conservation_score` table contains a BLOB field, which has been dumped as text.
This process - unless in very specific circumstances - is not reversible, and data cannot be loaded back again ... unless not using "LOAD DATA INFILE" naively.
Zeroing these data fixed the issue ... and luckily no test case were affected.

No changes to the test cases.

The test suite ran fine both on Codon and TravisCI.

### Changelog

N/A


TamaraNaboulsi (approved):
Various changes for the travis test suite run, mainly bumping up versions for multiple tools, in addition to some fixes in the way repos are clones. Looks good and tests are succeeding.

PR #660, opened 2024-08-23
Patch test dbs for Release 114

nwillhoft:
Update the API version to 114 and add 114 db schema patch.


sgiorgetti (approved):
Schema version bumped up to 114
The patch line is added to the "meta.txt" files
Increments look sensible
The number of patched files match the previous one for 113.

LGTM

PR #667, opened 2024-10-17
[113] Update docs pointer to latest version

nwillhoft:
Update the REST version and update the docs pointer.


sgiorgetti (approved):
OK.
Travis' failing because of inaccurate config ... which requires branch naming convention to fix. Overruling it this time.

PR #670, opened 2024-12-11
Deprecate DisGeNET (release 114)

dglemos:
This is an update for release 114

### Description

Deprecating the VEP plugin: DisGeNET

### Use case

The DisGeNET data is not freely available anymore.
The plugin does not make sense without the data.

### Possible Drawbacks

Users won't be able to run the plugin in the future.

### Testing

_Have you added/modified unit tests to test the changes?_

_If so, do the tests pass/fail?_

_Have you run the entire test suite and no regression was detected?_
no regression detected

### Changelog

[/vep] DisGeNET plugin is deprecated
[/ga4gh/beacon/query] Beacon does not return DisGeNET scores anymore

PR #671, opened 2024-12-12
Deprecate plugin 114

dglemos:
Same as https://github.com/Ensembl/ensembl-rest/pull/670

**Description**

Deprecating the VEP plugin: DisGeNET

**Use case**

The DisGeNET data is not freely available anymore.
The plugin does not make sense without the data.

**Benefits**

**Possible Drawbacks**

Users won't be able to run the plugin in the future.

**Testing**

Have you added/modified unit tests to test the changes?

If so, do the tests pass/fail?

Have you run the entire test suite and no regression was detected?
no regression detected

**Changelog**

[/vep] DisGeNET plugin is deprecated
[/ga4gh/beacon/query] Beacon does not return DisGeNET scores anymore

### DIFF
--- a/lib/EnsEMBL/REST.pm
+++ b/lib/EnsEMBL/REST.pm
@@ -64,7 +64,7 @@ use Catalyst qw/
 /;
 
 
-our $VERSION = '15.9';
+our $VERSION = '15.8';
 
 # Configure the application.
 #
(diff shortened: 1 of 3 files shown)


nwillhoft (approved):
PR to deprecate plugin for release 114. Changes approved by Variation in #670 and all looks good.

PR #692, opened 2025-07-06
Add new transcript attrib to REST lookup

nwillhoft:
### Description

Make the new transcript attribute, "Ensembl canonical extended" available via REST (as requested by Havana).

### Use case

Lookup transcripts with this new attribute tag.

### Benefits

Extra transcript attribute data available to users.

### Possible Drawbacks

n/a

### Testing

Yes.

Yes - build for the PR is passing. Branch PR is failing because is expecting version 116 but it getting 115. Believe will be resolved once branching done.

As mentioned, tests are failing as expected release 116 but getting release 115.

### Changelog

`[/lookup]` Added the ability to look up new `ens_canon_extended` transcript attribute flag.


sgiorgetti (approved):
LGTM
The build fails at branch level because of the schema changes being applied to `main` at a later stage.

